PR #8, opened 2021-06-04
fix: sprint15 bugs

KaWaite:
Fixes:
* update project image upload to use asset modal
*  remove workspace name from project deletion modal
* project creation modal now clears after creation
* widget button link works now
* Stop project or workspace creation if no name


src/components/molecules/Common/plugin/builtin/widgets/menu.tsx
```diff
@@ -77,16 +77,20 @@ const Menu: WidgetComponent<Property, PluginProperty> = ({ property }) => {
...
         const camera =
           "buttonCamera" in b ? b.buttonCamera : "menuCamera" in b ? b.menuCamera : undefined;
         setCamera(camera);
+      } else {
+        let link = "buttonLink" in b ? b.buttonLink : "menuLink" in b ? b.menuLink : undefined;
+        const splitLink = link?.split("/");
+        if (splitLink?.[0] !== "http:" && splitLink?.[0] !== "https:") {
+          link = "https://" + link;
+        }
+        window.open(link, "_blank", "noopener");
```
HideBa, line 93:
```suggestion
        const link = "buttonLink" in b ? b.buttonLink : "menuLink" in b ? b.menuLink : undefined;
        if(!link)return;
        if (!/https?:\/\/[A-Za-z0=9_\-\.]*).test(link) {
          link = "https://" + link;
        }
        window.open(link, "_blank", "noopener");
```

HideBa: However, do we support the string without `http` or `https`? We never know if the linked website support TLS.

If we don't support

```
const link = "buttonLink" in b ? b.buttonLink : "menuLink" in b ? b.menuLink : undefined;
if(!link || !/https?:\/\/[A-Za-z0=9_\-\.]*).test(link))return;
window.open(link, "_blank", "noopener");
```

is also fine.

HideBa (changes requested):
Thanks for quickly fixing bugs.

src/components/molecules/Common/plugin/builtin/widgets/menu.tsx
KaWaite: I did this because without the full url window.open will append the url to the current location. like localhost/www.google.ca

rot1024 (approved):

HideBa (approved):

PR #11, opened 2021-06-07
fix: menu widget & change asset modal default order

KaWaite:
Bug fix:
- menu widget's item functionality wasn't working (link and flight)

Change:
- Asset modal now shows newest items at the top by default


HideBa (approved):
LGTM! Thanks for fixing quickly. 👍

PR #14, opened 2021-06-13
feat: import google sheet dataset

issmail-basel:
- Add gapi (google API) script when initializing the component
- Initializing both (auth and file picker)
- Put the right permissions
- Add required modals
- Call backend mutation for importing datasheets
**Screenshots**


src/components/atoms/Icon/Icons/sheet-file.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="22" height="20" viewBox="0 0 22 20" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M1.34343 5.8147H5.66213C6.40272 5.8147 7.00509 5.21902 7.00509 4.48656V1.32837C7.00509 0.595913 6.40248 0 5.66165 0H1.34296C0.60237 0 0 0.595674 0 1.32789V4.48633C0 5.21878 0.602609 5.8147 1.34343 5.8147ZM5.09209 3.9017H1.91309V1.91333H5.09209V3.9017Z" fill="#C7C5C5"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M10.006 5.8147H20.698C21.4158 5.8147 22 5.23313 22 4.51837V1.29633C22 0.581565 21.4158 0 20.6977 0H10.0055C9.28763 0 8.70343 0.581565 8.70343 1.29633V4.51813C8.70343 5.23313 9.28763 5.8147 10.006 5.8147ZM20.0869 3.9017H10.6164V1.91333H20.0869V3.9017Z" fill="#C7C5C5"/>
```
KaWaite, line 3: I mentioned before, but we need all svg's fill values to be `currentColor` so that we can easily change the color from within the Icon component.

src/components/molecules/Common/AssetModal/AssetListItem/index.tsx
```diff
@@ -10,11 +10,11 @@ import { metricsSizes } from "@reearth/theme/metrics";
...
   id: string;
-  teamId: string;
+  teamId?: string;
   name: string;
-  size: number;
-  url: string;
-  contentType: string;
+  size?: number;
+  url?: string;
+  contentType?: string;
```
KaWaite, line 17: Any reason to make teamId, size, url and contentType optional connected with google sheets? The Asset type in Golang and GraphQL expect all of these elements so we don't want to make them optional.

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/Gdrive/hooks.ts
```diff
@@ -0,0 +1,136 @@
...
+declare const gapi: any;
+declare const google: any;
+
+export type SheetParameter = {
+  accessToken: string;
+  fileId: string;
+  sheetName: string;
+};
+
+export type GoolgeSheet = {
```
KaWaite, line 12: GoolgeSheet should be GoogleSheet

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/Gdrive/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+    selectedSheet,
+    handleAuthClick,
+    handleSheetSelect,
+  } = useHooks(onSheetSelect);
+
+  return (
+    <>
+      <StyledIcon icon={"arrowLongLeft"} size={24} onClick={onReturn} />
+      {syncLoading && <Loading />}
+      {pickedFile?.id && (
```
KaWaite, line 39: Can you move the Flex and Divider outside of both conditional areas? Both areas (whether pickedFile?.id or !pickedFile?.id) include the first Flex container and its contents + Divider so I don't see why they can't be moved outside to always be shown.

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/Gdrive/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+            <Text
+              size="m"
+              color={theme.colors.text.weak}
+              otherProperties={{ marginBottom: metricsSizes["m"] + "px" }}>
+              {intl.formatMessage({
+                defaultMessage: "Re:earth supports Google Sheets, and CSV file to upload.",
+              })}
+            </Text>
+          </Flex>
+          <Flex justify="center">
```
KaWaite, line 100: Is there any way to move this Flex's contents into the Flex just above? Flex is just a div with flexbox so there isn't REALLY any semantic reason to split the contents unless they have different styling needs. 
Actually, instead of the above suggestion can you not remove the Flex above wrapping the single Text component and just do margin left and right auto to the Text component itself? That way we can keep the DOM tree from getting too out of hand.

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/Gdrive/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+const GdriveIcon = styled(Icon)`
+  margin-right: ${metricsSizes["m"]}px;
+`;
+
+const StyledIcon = styled(Icon)`
+  cursor: pointer;
+`;
+
+const AssetList = styled(Flex)`
+  flex-direction: column;
```
KaWaite, line 127: Just to keep things clean if you can move this into the component's props with the other flex related attributes that will keep things easier to change/troubleshoot later

src/components/organisms/EarthEditor/DataSourcePane/hooks.ts
```diff
@@ -92,6 +93,26 @@ export default () => {
...
+        variables: {
+          accessToken,
+          fileId,
+          sheetName,
+          sceneId,
+          datasetSchemaId: schemeId,
+        },
+      });
+      // re-render
+      // await client.resetStore();
```
KaWaite, line 111: Is this commented out code forgotten or on purpose?

KaWaite (commented):
Left some comments. Let me know if you have any questions or when you are ready for my re-review!

src/components/molecules/Common/AssetModal/AssetListItem/index.tsx
issmail-basel: In design I should use AssetList.
But google sheet API (for single sheets) does not provide any of these data (size, contentType) and also URL is not needed.

KaWaite: Ah I see, gotcha!

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/Gdrive/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        )}
+
+        {!pickedFile?.id && (
+          <>
+            <Text
+              size="m"
+              color={theme.colors.text.weak}
+              otherProperties={{ marginBottom: metricsSizes["m"] + "px" }}>
+              {intl.formatMessage({
+                defaultMessage: "Re:earth supports Google Sheets, and CSV file to upload.",
```
KaWaite, line 87: Sorry missed this. Should be "Re:Earth supports uploading Google Sheets and CSV files."

KaWaite (dismissed):
Please change where I made a comment about English.
But other than that looks good to me!

HideBa (dismissed):
This PR looks good to me. 👍  However, the `AssetModal` had been changed somehow. Asset and Dataset is a different thing for Re:Earth. As a refactoring task, AssetModal can be renamed `GridModal` or `CardModal` and made a generic component which accepts `Asset` type and `Dataset` type.

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -10,7 +10,10 @@ type CardSize = "small" | "medium" | "large";
 export type Props = {
   className?: string;
   name: string;
-  url: string;
+  url?: string;
+  icon?: string;
+  iconSize?: string;
+  isImage?: boolean;
```
KaWaite, line 16: Do you need the isImage? If you don't, it must have crept back during your merge.
@lavalse removed it in favor of the local regex expression since I believe the AssetContainer (which uses the  AssetCard) is being used now in the Settings page with ALL assets coming in. Which means each url needs to be checked locally to assign an icon or not.
Sorry, I know it might mean a bit of re-refactoring.

lavalse: yes, sorry for removing "isImage". but which is necessary , like the asset is image or not image , but they will be show in setting page at same time. if any difficult let me know

KaWaite (commented):
Sorry looks like you mixed current code with old code when you merged. Please check the comment I left

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/hooks.ts
```diff
@@ -2,28 +2,44 @@ import { useCallback, useState, useEffect } from "react";
 import { useIntl } from "react-intl";
 import useFileInput from "use-file-input";
 import { Type as NotificationType } from "@reearth/components/atoms/NotificationBar";
+import { SheetParameter } from "./Gdrive/hooks";
```
rot1024, line 5: Re-export SheetParameter at `./Gdrive/index.tsx` for SheetParameter to enable to be imported from `./Gdrive`.

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/Gdrive/hooks.ts
```diff
@@ -0,0 +1,136 @@
+import { useCallback, useEffect, useState } from "react";
+
+declare const gapi: any;
+declare const google: any;
```
rot1024, line 4: https://www.npmjs.com/package/@types/google.picker is usable

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/index.tsx
```diff
@@ -81,11 +95,24 @@ const DatasetModal: React.FC<Props> = ({
...
+            <Card
+              id="gdrive"
+              icon="googleDrive"
+              iconSize="50px"
+              text={intl.formatMessage({ defaultMessage: "Google Drive" })}
+              margin={56}
+              border="dashed"
+              borderColor={colors.outline.weak}
+              onClick={handleClick}
+            />
```
rot1024, line 107: Hide Google Drive when there is no Google API key in REEARTH_CONFIG.

rot1024 (changes requested):
Have you considered using https://www.npmjs.com/package/react-google-picker  ?

issmail-basel:
@rot1024 , yes I've considered using  https://www.npmjs.com/package/react-google-picker 
but we only use this picker once, and in a very simple way, so I guess no need for an external package.
If there is any future need, I'll consider changing my code

rot1024 (approved):

HideBa (dismissed):

PR #15, opened 2021-06-15
fix: force logout when me query returns null

HideBa:
## Before
If the `me` query returns `null`. The screen becomes black and the user couldn't do anything.

## After
Without the valid query result, force the user to log out.

### DIFF
--- a/src/components/pages/TopPage/hooks.ts
+++ b/src/components/pages/TopPage/hooks.ts
@@ -24,8 +24,7 @@ export default () => {
   }, [isAuthenticated, navigate, currentTeam, setLocalState, data, teamId]);
 
   useEffect(() => {
-    if (authError || (!loading && !data?.me === null)) {
-      console.log("logout--------");
+    if (authError || (!loading && data?.me === null)) {
       logout();
     }
   }, [authError, data?.me, loading, logout]);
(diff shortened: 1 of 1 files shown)


KaWaite (approved):

src/components/pages/TopPage/hooks.ts
```diff
@@ -24,10 +24,10 @@ export default () => {
   }, [isAuthenticated, navigate, currentTeam, setLocalState, data, teamId]);
 
   useEffect(() => {
-    if (authError) {
+    if (authError || (!loading && data?.me === null)) {
```
rot1024, line 27:
```suggestion
    if (authError || (!loading && !data?.me)) {
```

src/components/pages/TopPage/hooks.ts
```diff
@@ -24,10 +24,10 @@ export default () => {
...
       logout();
```
rot1024, line 28: In the future, I would like to display an error message after logging out, because currently it logs out silently.

rot1024 (approved):

rot1024 (commented):
fix: force logout without me -> fix: force logout when me query returns null now

Changed the title for better clarity.

PR #16, opened 2021-06-15
fix: layer panel bug

KaWaite:
- fix bug where deselecting a child layer would close the folder
-  update to show indicator beside parent when child is selected
Screen Shot 2021-06-15 at 15 49 00


rot1024 (approved):
OK, but the title of the pull request should be more descriptive 👍

HideBa (approved):

PR #17, opened 2021-06-17
fix: infinit logout loop

HideBa:


src/components/pages/TopPage/hooks.ts
```diff
@@ -24,7 +24,7 @@ export default () => {
   }, [isAuthenticated, navigate, currentTeam, setLocalState, data, teamId]);
 
   useEffect(() => {
-    if (authError || (!loading && !data?.me)) {
+    if (authError || (!loading && data?.me === null)) {
```
KaWaite, line 27: I am not sure if this will call before or after the query starts so is it better (like @rot1024 suggested) to make it `(isAuthenticated &&  !loading && data?.me === null)`.

HideBa: Thanks! I fixed it.

rot1024 (approved):

KaWaite (approved):

PR #18, opened 2021-06-17
chore: set up sentry

HideBa:
## What 
* Added sentry settings
* Added apollo link to add sentry's breadcrumb which helps us to investigate the errors

## Why
Even if the user gets errors, we didn't know that. To catch up what kind of errors our user has, I added Sentry which is error tracking tool.


src/gql/provider.tsx
```diff
@@ -7,6 +7,8 @@ import { useAuth0 } from "@auth0/auth0-react";
 
 import { store, localSlice } from "@reearth/state";
 import fragmentMatcher from "./fragmentMatcher.json";
+import { SentryLink } from "apollo-link-sentry";
+import { reportError } from "@reearth/sentry";
```
rot1024, line 11: Please note the order of import:

1. External modules
2. `@reearth/...`
3. Relative module (children and siblings)

src/gql/provider.tsx
```diff
@@ -34,8 +36,11 @@ const Provider: React.FC = ({ children }) => {
     if (!networkError && !graphQLErrors) return;
     const error = networkError?.message ?? graphQLErrors?.map(e => e.message).join(", ");
     store.dispatch(localSlice.actions.set({ error }));
+    error && reportError(error);
```
rot1024, line 39:
```suggestion
    if (error) reportError(error);
```

Please use `&&` instead of `if` statement only in JSX.

src/sentry/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+};
+
+export const reportError = (error: string) => {
+  Sentry.captureException(error);
+};
+
+export default {
+  reportError,
+};
```
rot1024, line 20: You don't need to create a directory, the file can simply be `src/sentry.ts`.

src/sentry/index.ts
```diff
@@ -0,0 +1,20 @@
+import * as Sentry from "@sentry/browser";
+
+export const initializeSentry = () => {
```
rot1024, line 3: In the import statement, it is clear that this function is from sentry, as shown below, so the name of the function can be concise.

```ts
import { initialize as initializeSentry } from "./sentry";
```

```suggestion
export const initialize = () => {
```

src/sentry/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  const sentryDSN = window.REEARTH_CONFIG?.sentryDsn;
+  const sentryEnv = window.REEARTH_CONFIG?.sentryEnv;
```
rot1024, line 5:
```suggestion
  const { sentryDSN, sentryEnv } = window.REEARTH_CONFIG ?? {};
```

src/sentry/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (sentryDSN) {
+    Sentry.init({
+      dsn: sentryDSN,
+      environment: sentryEnv,
+    });
+  }
+};
+
+export const reportError = (error: string) => {
+  Sentry.captureException(error);
```
rot1024, line 15: Have you verified that this function does not generate an error even if sentry is not initialized?

rot1024 (changes requested):

issmail-basel (approved):
Other than @rot1024 comments, seems good.

src/sentry/index.ts
HideBa: OK! Thank you.

src/sentry/index.ts
HideBa: I've tested it if it works without initilization. 👍

HideBa:
@rot1024 I've fixed where you had reviewed. Plz review again.

src/sentry.ts
```diff
@@ -0,0 +1,15 @@
...
+  const { sentryDsn, sentryEnv } = window.REEARTH_CONFIG ?? {};
+  if (sentryDsn) {
+    Sentry.init({
+      dsn: sentryDsn,
+      environment: sentryEnv,
+    });
+  }
+};
+
+export const reportError = (error: string) => {
```
rot1024, line 13:
```suggestion
export const reportError = (error: any) => {
```

Type of `error` should be `any` because `Sentry.captureException` accepts `exception` arg with `any`.

https://github.com/getsentry/sentry-javascript/blob/67b76854bc0cee526930780e1d22f67dec1ff2bf/packages/core/src/baseclient.ts#L100

rot1024 (approved):

src/gql/provider.tsx
```diff
@@ -51,7 +56,7 @@ const Provider: React.FC = ({ children }) => {
 
   const client = new ApolloClient({
     uri: endpoint,
-    link: ApolloLink.from([errorLink, authLink, uploadLink]),
+    link: ApolloLink.from([sentryLink, errorLink, authLink, uploadLink]),
```
rot1024, line 59:
```suggestion
    link: ApolloLink.from([errorLink, sentryLink, authLink, uploadLink]),
```

https://github.com/DiederikvandenB/apollo-link-sentry#faq

> Note that this package (currently) only adds breadcrumbs. This means that you are still responsible for reporting errors to Sentry.

So it is desirable to actually send the error to Sentry only after it has passed through the SentryLink process.

Process flow:

```
errorLink -> sentryLink -> authLink -> uploadLink -> (sending query and error happen) -> uploadLink -> authLink -> sentryLink (attach breadcrumbs) -> errorLink
```

rot1024 (changes requested):

src/gql/provider.tsx
HideBa: Thank you!

rot1024 (approved):

PR #19, opened 2021-06-18
feat: refine setting page

lavalse:
- update the setting page with new design ( new narvigation with responsive react, new structure )
- update asset library page with Assets Container component, and fix bugs
- update projects workspaces list page
- update properties filed style


src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -35,8 +33,12 @@ const AssetCard: React.FC<Props> = ({
       selected={selected}
       cardSize={cardSize}
       onClick={() => onCheck?.(!check)}>
-      <ImgWrapper cardSize={cardSize} url={url}>
-        {!isImage && <Icon icon="file" />}
+      <ImgWrapper cardSize={cardSize}>
+        {/\.(jpg|jpeg|png|gif|GIF|JPG|PNG)$/.test(url) ? (
```
rot1024, line 37:
```suggestion
        {/\.(jpg|jpeg|png|gif|svg|webp|GIF|JPG|JEPG|PNG|SVG|WEBP)$/.test(url) ? (
```

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -57,16 +59,18 @@ const Wrapper = styled(Flex)<{ selected?: boolean; cardSize?: CardSize }>`
   box-shadow: 0 6px 6px -6px ${props => props.theme.colors.other.black};
   border: 1px solid
     ${props => (props.selected ? `${props.theme.assetCard.highlight}` : "transparent")};
-  margin: ${({ cardSize }) => (cardSize === "small" ? "5px" : "10px")};
+  /* margin: ${({ cardSize }) => (cardSize === "small" ? "5px" : "10px")}; */
```
rot1024, line 62:
```suggestion
```

Please remove the line if it is no longer used.

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -13,8 +13,10 @@ import { styled } from "@reearth/theme";
 import AssetCard from "../AssetCard";
 import AssetListItem from "../AssetListItem";
 import AssetSelect from "../AssetSelect";
+import VirtualAssetCard from "../VirtualAssetCard";
 
 import useHooks, { Asset as AssetType, LayoutTypes, FilterTypes } from "./hooks";
+import { useCallback, useRef, useState, useEffect } from "react";
```
rot1024, line 19: Please move `import {...} from React` to the first line.

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -70,20 +76,55 @@ const AssetContainer: React.FC<Props> = ({
...
+    }
+  }, [containerWidth]);
+
+  const handleResize = () => {
+    if (listRef.current?.offsetWidth) {
+      setContainerWidth(listRef.current?.offsetWidth);
+    }
+  };
+
+  window.addEventListener("resize", handleResize);
```
rot1024, line 100: use `useEffect` to attach a event in React components. Otherwise, each time the component is rendered, a new callback will be registered.

```ts
useEffect(() => {
  const handleResize = () => {
    if (listRef.current?.offsetWidth) {      
      setContainerWidth(listRef.current?.offsetWidth);
    }
  };
  window.addEventListener("resize", handleResize);
  return () => window.removeEventListener("resize", handleResize);
}, []);
```

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -166,35 +210,35 @@ const AssetContainer: React.FC<Props> = ({
...
+const ButtonWrapper = styled.div<{ onRemove?: (assets: AssetType[]) => void }>`
+  display: flex;
+  justify-content: ${({ onRemove }) => (onRemove ? "flex-end" : "center")};
 `;
 
 const AssetList = styled(Flex)<{ layoutType?: LayoutTypes }>`
-  ${({ layoutType }) => layoutType === "list" && "flex-direction: column;"}
-  max-height: 458px;
+  ${({ layoutType }) => layoutType === "list" && "flex-direction: column;"};
+  // max-height: 458px;
```
rot1024, line 231:
```suggestion
```

Don't leave any lines of code that are simply commented out. Since you're using git, you can always revert it.

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        }
+      }
+    }
+  }, [assetsLength, cardSize, containerWidth]);
+
+  useEffect(() => {
+    handleVirtualWidth();
+  }, [containerWidth, handleVirtualWidth]);
+
+  return <Wrapper width={virtualWidth} cardSize={cardSize}></Wrapper>;
```
rot1024, line 37:
```suggestion
  return <Wrapper width={virtualWidth} cardSize={cardSize} />;
```

src/components/molecules/Settings/Account/AccountSection/index.tsx
```diff
@@ -2,10 +2,10 @@ import React, { useEffect, useState } from "react";
 import PasswordModal from "@reearth/components/molecules/Settings/Account/PasswordModal";
 import Section from "@reearth/components/molecules/Settings/Section";
 import EditableItem from "@reearth/components/molecules/Settings/Project/EditableItem";
+import Field from "@reearth/components/molecules/Settings/Field";
 import Icon from "@reearth/components/atoms/Icon";
-import { styled, useTheme } from "@reearth/theme";
+import { styled } from "@reearth/theme";
 import { useIntl } from "react-intl";
-import Text from "@reearth/components/atoms/Text";
```
rot1024, line 8: Please note import order:

1. external modules (react, ...)
2. reearth modules (`@reearth/...`)
3. internal modules (relative path)

translations/ja.yml
```diff
@@ -352,7 +357,7 @@ src:
           '3590336714': エクスポート
       Settings:
         Account:
-          '3198092477': アカウント
+          '3814251771': 'マイ アカウント'
```
rot1024, line 360: 細かいかもしれませんが、日本語ではあまり「マイ アカウント」という表記をUIでは見かけないと思います。「アカウント」のままでも良いでしょう。

src/components/organisms/Settings/Workspace/Asset/index.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import { useIntl } from "react-intl";
 import useHooks from "./hooks";
 import SettingPage from "@reearth/components/organisms/Settings/SettingPage";
-import AssetSection from "@reearth/components/molecules/Settings/Workspace/Asset/AssetSection";
+import AssetSection from "@reearth/components/molecules/Settings/Workspace/AssetSection";
```
rot1024, line 5: Can you explain why AssetSection has been moved to Workspace?

AssetSection is used only inside Asset, so I think it should be in Asset directory.

src/components/organisms/Settings/ProjectList/hooks.ts
```diff
@@ -28,9 +28,7 @@ export default () => {
   const openModal = useCallback(() => setModalShown(true), []);
 
   const { data, loading, refetch } = useMeQuery();
-  const [createNewProject] = useCreateProjectMutation({
-    refetchQueries: ["Project"],
-  });
+  const [createNewProject] = useCreateProjectMutation();
```
rot1024, line 31: You are turning off refetch, did you have any problems with it?

src/components/organisms/EarthEditor/PropertyPane/hooks.ts
```diff
@@ -184,13 +184,14 @@ export default (mode: Mode) => {
...
   const createAssets = useCallback(
-    (file: File) =>
+    (files: FileList) =>
       (async () => {
         if (teamId) {
-          await createAssetMutation({
-            variables: { teamId, file },
-            refetchQueries: ["Assets"],
-          });
+          await Promise.all(
```
rot1024, line 190:
```suggestion
　　　　　　// TODO: support multiple file uploads
          await Promise.all(
```

src/components/molecules/Settings/Workspace/AssetSection/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  name: string;
+  size: number;
+  url: string;
+  contentType: string;
+};
+
+type Props = {
+  assets?: Asset[];
+  onCreate?: (files: FileList) => void;
+  onRemove?: (assets: Asset[]) => void;
```
rot1024, line 16:
```suggestion
  onRemove?: (assets: string[]) => void;
```

I think the ID is enough to identify the target of deletion.

src/components/molecules/Settings/SettingPage/index.tsx
```diff
@@ -20,9 +20,14 @@ const SettingPage: React.FC<Props> = ({
     <ProjectMenu currentProject={currentProject} teamId={currentTeam?.id} />
   );
 
+  const [isOpen, setIsOpen] = useState(false);
+  const handleClick = () => {
+    setIsOpen(!isOpen);
```
rot1024, line 25:
```suggestion
    setIsOpen(o => !o);
```

src/components/molecules/Settings/NavigationItem/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import { styled, fonts } from "@reearth/theme";
 import { Link } from "@reach/router";
+import colors from "@reearth/theme/colors";
```
rot1024, line 4:
```suggestion
import { colors } from "@reearth/theme";
```

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -23,22 +25,26 @@ export type Props = {
...
   assets,
   isMultipleSelectable = false,
   accept,
   onCreateAsset,
+  onRemove,
   initialAsset,
   selectedAssets,
   selectAsset,
   fileType,
+  isSettingPage,
```
rot1024, line 47:
```suggestion
  isHeightFxied,
```

isSettingPage is fine, but it is not versatile. If you use a name like this, it will be convenient when you want to use this property outside of SettingPage.

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -23,22 +25,26 @@ export type Props = {
   assets?: Asset[];
   isMultipleSelectable?: boolean;
   accept?: string;
-  onCreateAsset?: (file: File) => void;
+  onCreateAsset?: (files: FileList) => void;
+  onRemove?: (assets: AssetType[]) => void;
```
rot1024, line 29:
```suggestion
  onRemove?: (assets: string[]) => void;
```

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -70,20 +76,55 @@ const AssetContainer: React.FC<Props> = ({
     { key: "name", label: intl.formatMessage({ defaultMessage: "Alphabetical" }) },
   ];
 
+  const handleRemove = useCallback(() => {
+    if (selectedAssets?.length) {
+      onRemove?.(selectedAssets);
```
rot1024, line 81:
```suggestion
      onRemove?.(selectedAssets.map(a => a.id));
```

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  cardSize?: CardSize;
+};
+
+const VirtualAssetCard: React.FC<Props> = ({ assetsLength, containerWidth, cardSize }) => {
+  const [virtualWidth, setVirtualWidth] = useState(0);
+
+  const handleVirtualWidth = useCallback(() => {
+    if (containerWidth) {
+      if (assetsLength) {
+        const cardWidth = cardSize === "medium" ? 216 : 128;
```
rot1024, line 18: I'm a little concerned that the magic numbers (216 and 128) is hard-coded this way. It is easier to understand if you define it as a const outside the component.

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+          setVirtualWidth(0);
+        } else {
+          setVirtualWidth(virtualBoxWidth);
+        }
+      }
+    }
+  }, [assetsLength, cardSize, containerWidth]);
+
+  useEffect(() => {
+    handleVirtualWidth();
```
rot1024, line 34: You don't need to use useCallback, you can just write the contents of the function here.

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        const gapWidth = (containerWidth - cardWidth * numInrow) / (numInrow - 1);
+        const virtualBoxNumber = numInrow - lastRowNum;
+        const virtualBoxWidth = virtualBoxNumber * cardWidth + gapWidth * (virtualBoxNumber - 1);
+        if (lastRowNum === 0) {
+          setVirtualWidth(0);
+        } else {
+          setVirtualWidth(virtualBoxWidth);
+        }
+      }
+    }
```
rot1024, line 30: If you can explain in the comments why you want to do this kind of calculation, the code will be understandable for others to read. 

Also, have you considered whether this can be solved with CSS?

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -70,20 +76,55 @@ const AssetContainer: React.FC<Props> = ({
...
+  }, [onRemove, selectedAssets]);
+
+  const listRef = useRef<HTMLDivElement>(null);
+  const [containerWidth, setContainerWidth] = useState(0);
+
+  useEffect(() => {
+    if (listRef.current?.offsetWidth) {
+      setContainerWidth(listRef.current?.offsetWidth);
+    }
+  }, [containerWidth]);
```
rot1024, line 92: This useEffect can cause an infinite loop. (containerWidth update -> setContainerWidth -> containerWidth update -> ...)

rot1024 (changes requested):

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
lavalse: sorry, i missed it

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: yes, I hope a new callback be registered everytime the component is rendered. because it need a initialization Calculation for the size... so... any suggestion? @rot1024

rot1024: The code I've shown you should work just fine.

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
lavalse: So you suggest to write some Code remarks here ? sure
I want to create a virtual object which takes the space inside the flex box, which can align the other objects.
yes, I have considered the css way, but I think this is the best way to sovle this problem, any suggetions ?

rot1024: What features do you need it for?
Maybe we can use CSS Grid or something?

src/components/molecules/Settings/Account/AccountSection/index.tsx
lavalse: ok, thanks.

src/components/organisms/Settings/Workspace/Asset/index.tsx
lavalse: yes, that's true. I move this asset part from project, I guess I see the dir path look like this , which is ok.
image
any way, let me change it

src/components/organisms/Settings/ProjectList/hooks.ts
lavalse: Beacuse it refect below inside the createProject function, and also it didn't work here.

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: ok, thanks

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
lavalse: sure, thanks

src/components/atoms/PublicationStatus/index.tsx
```diff
@@ -36,7 +37,7 @@ const PublicationStatus: React.FC<PublishStatusProps> = ({
       <Text
         size={size === "lg" ? "m" : size === "md" ? "s" : "xs"}
         otherProperties={{ userSelect: "none" }}
-        customColor>
+        color={color}>
```
KaWaite, line 40: I see that the color prop wasn't being used, so passing it here is fine, but did you make sure this PublicationStatus component being used elsewhere isn't making use of the customColor (aka inheriting a color from a parent component)?

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
KaWaite: Isn't this a typo? JEPG -> JPEG

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -35,8 +33,12 @@ const AssetCard: React.FC<Props> = ({
       selected={selected}
       cardSize={cardSize}
       onClick={() => onCheck?.(!check)}>
-      <ImgWrapper cardSize={cardSize} url={url}>
-        {!isImage && <Icon icon="file" />}
+      <ImgWrapper cardSize={cardSize}>
+        {/\.(jpg|jpeg|png|gif|svg|webp|GIF|JPG|JEPG|PNG|SVG|WEBP)$/.test(url) ? (
+          <PreviewImage url={url}></PreviewImage>
```
KaWaite, line 38: Unless there is a JEPG image format I am unaware of that is a typo ;) Should be JPEG, right?

lavalse: yesyes, sorry, I just copy from inoue-san, thanks

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -57,16 +59,17 @@ const Wrapper = styled(Flex)<{ selected?: boolean; cardSize?: CardSize }>`
...
-  margin: ${({ cardSize }) => (cardSize === "small" ? "5px" : "10px")};
+  margin: 12px;
   padding: ${({ cardSize }) =>
     cardSize === "small" ? "8px" : cardSize === "medium" ? "12px" : "20px"};
   width: ${({ cardSize }) =>
-    cardSize === "small" ? "104px" : cardSize === "medium" ? "163px" : "218px"};
+    cardSize === "small" ? "104px" : cardSize === "medium" ? "192px" : "274px"};
   height: ${({ cardSize }) =>
-    cardSize === "small" ? "104px" : cardSize === "medium" ? "168px" : "234px"};
+    cardSize === "small" ? "104px" : cardSize === "medium" ? "186px" : "257px"};
```
KaWaite, line 68: I am sure you have, but did you make sure the styles are still okay? I am pretty sure the values you added are the values in Figma, but the reason I initially reduced them was taking into consideration the padding + any discrepancies because of borders,etc that might be adding width or height unexpectedly(compared to the Figma design).

lavalse: yes, yes, don't worry. I made them box-sizing. and adjuest them

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
KaWaite: Also if PreviewImage is getting no children, just make it
`<PreviewImage url={url} />`

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -166,35 +209,34 @@ const AssetContainer: React.FC<Props> = ({
 };
 
 const Wrapper = styled.div`
-  height: 558px;
   width: 100%;
 `;
 
-const AssetWrapper = styled(Flex)`
-  height: 425px;
+const AssetWrapper = styled.div<{ isHeightFixed?: boolean }>`
```
KaWaite, line 215: Why did you turn this into a div and then place flex on it? It was originally using our Flex component. Any limitations you found on the Flex component that made you switch to a div?

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -166,35 +209,34 @@ const AssetContainer: React.FC<Props> = ({
...
+  height: ${({ isHeightFixed }) => (isHeightFixed ? "" : "425px")};
+  min-height: 400px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+`;
+
+const ButtonWrapper = styled.div<{ onRemove?: (assets: AssetType[]) => void }>`
```
KaWaite, line 223: Any reason this isn't a Flex component? If it is a Flex component you don't need to make a styled component here. Then just do something like:
`<Flex justify={onRemove ? "flex......>....</Flex>`

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -166,35 +209,34 @@ const AssetContainer: React.FC<Props> = ({
...
+  display: flex;
+  justify-content: ${({ onRemove }) => (onRemove ? "flex-end" : "center")};
 `;
 
 const AssetList = styled(Flex)<{ layoutType?: LayoutTypes }>`
-  ${({ layoutType }) => layoutType === "list" && "flex-direction: column;"}
-  max-height: 458px;
+  ${({ layoutType }) => layoutType === "list" && "flex-direction: column;"};
+
```
KaWaite, line 230: Be careful of that extra ";".  I am not sure it is fine in styled components, but that might be shown somewhere on the webpage :P As you can see, the conditional flex-direction has a ";" already because if layoutType is NOT a list, there is no flex-direction.
Also, becuase this is a Flex component you can (and should) put this up in the component itself (like `<Flex direction={layoutType === "list" ? "column" : undefined}>` ***Just double check that undefined works. Might need to be null***

src/components/molecules/Common/AssetModal/AssetListItem/index.tsx
```diff
@@ -25,12 +25,18 @@ export type Props = {
...
+const AssetListItem: React.FC<Props> = ({ asset, selected, checked, onCheck }) => {
   const theme = useTheme();
   return (
     <ListItem key={asset.id} align="center" selected={selected} onClick={() => onCheck?.(!check)}>
       <Icon
-        icon={checked ? "checkCircle" : isImage ? "image" : "file"}
+        icon={
+          checked
+            ? "checkCircle"
+            : /\.(jpg|jpeg|png|gif|GIF|JPG|PNG)$/.test(asset.url)
```
KaWaite, line 36: Here needs to be updated like in the AssetCard. I am sure isImage isn't the greatest, but I am not sure the same regex in multiple areas is good either.
Any big reason you removed the isImage?

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  useEffect(() => {
+    handleVirtualWidth();
+  }, [containerWidth, handleVirtualWidth]);
+
+  return <Wrapper width={virtualWidth} cardSize={cardSize} />;
+};
+
+const Wrapper = styled.div<{ width?: number; cardSize?: CardSize }>`
+  width: ${p => p.width}px;
```
KaWaite, line 43: Small thing, but to try and keep all code as similar stylistically as possible, can you change this to either 
`props => props.width`
or
`${({width}) => width}px`

src/components/molecules/Common/Header/index.tsx
```diff
@@ -105,15 +105,19 @@ const Header: React.FC<Props> = ({
 
 const Wrapper = styled.header`
   width: 100%;
-  height: ${metrics.headerHeight}px;
+  min-height: ${metrics.headerHeight}px;
```
KaWaite, line 108: With this changed to min-height, does the header  grow its height at any point as the width decreases because of children bunching up?

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
lavalse: ok, thanks

src/components/molecules/Common/Header/index.tsx
lavalse: yes, sometimes when the project name is a long name. when screen size became smaller, the project name will jump out of the header bar, witch is quite terrible. So we did nothong to handle the project name length for now, I think it's better to not limite the header height, so the project name will not jump out of header

src/components/molecules/Common/AssetModal/AssetListItem/index.tsx
lavalse: yes, sure. I m not update here yet. thanks

src/components/molecules/Settings/Account/AccountSection/index.tsx
```diff
@@ -72,31 +64,15 @@ const ProfileSection: React.FC<Props> = ({
...
-  display: inline-block;
 `;
 
 const StyledIcon = styled(Icon)`
-  display: inline;
-  float: right;
   cursor: pointer;
+
+  &:hover {
+    color: ${props => props.theme.colors.text.strong};
```
KaWaite, line 74: If you are using colors directly, I believe you can import it above with styled and then pass it in like:
`${({colors}) => colors.blahblah}`

src/components/molecules/Settings/Field/index.tsx
```diff
@@ -32,34 +32,34 @@ const Field: React.FC<Props> = ({ className, header, action, secondaryAction, bo
...
   flex-direction: row;
-  justify-content: center;
-  width: 100%;
+  justify-content: space-between;
+  justify-items: center;
   color: ${colors.text.main};
 
   &:not(:last-child) {
-    margin-bottom: 40px;
+    margin-bottom: 24px;
```
KaWaite, line 40: In the design metrics isn't there a 24px value (maybe xl or 2xl?)??

src/components/molecules/Settings/Field/index.tsx
```diff
@@ -32,34 +32,34 @@ const Field: React.FC<Props> = ({ className, header, action, secondaryAction, bo
...
 `;
 
 const Content = styled.div`
-  flex-grow: 1;
+  max-height: 264px;
+  overflow: auto;
+  flex: 1;
+`;
+
+const ActionArea = styled.div`
```
KaWaite, line 54: This could be a Flex component and you wouldn't need to create a styled component

src/components/molecules/Common/AssetModal/AssetListItem/index.tsx
lavalse: yes , why I remove is because the Assets container in the setting page, there are not only image type, and also kml type and others. and also the image type should show the preview image, but the kml file don't need. So the card should decide by themselves, not get the isImage from outside. So I think this way is better @KaWaite

src/components/molecules/Settings/Field/index.tsx
KaWaite: Also don't be afraid to updated components if they could be Flex (or Text or another one of our atoms components). There are some stragglers from before we had those components!

src/components/molecules/Settings/Logo/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import Icon from "@reearth/components/atoms/Icon";
+import { styled } from "@reearth/theme";
+
+export interface Props {
+  className?: string;
+}
+
+const Logo: React.FC<Props> = ({ className }) => {
```
KaWaite, line 9: Is this component used in multiple places? If not, I am not sure it is necessary to make this its own component. Let me know what you think

src/components/molecules/Settings/Navigation/index.tsx
```diff
@@ -66,21 +70,25 @@ const Navigation: React.FC<Props> = ({ team, project }) => {
...
 const NavigationList = styled.ul`
   width: 100%;
   margin: 0;
   padding: 0;
-  font-size: ${fonts.sizes.l}px;
-  font-weight: bold;
+  font-size: ${fonts.sizes.m}px;
+`;
+
+const Line = styled.hr`
```
KaWaite, line 89: We have a Divider atom component. If that is useable here think about using that ;)

src/components/molecules/Settings/NavigationItem/index.tsx
```diff
@@ -22,22 +27,18 @@ const NavigationItem: React.FC<Props> = ({ name, to, children }) => {
 
 const LinkItem = styled(Link)`
   display: flex;
-  padding: 16px;
-  color: #ffffff;
+  padding: 16px 8px;
```
KaWaite, line 30: metrics?

src/components/molecules/Settings/NavigationItem/index.tsx
```diff
@@ -22,22 +27,18 @@ const NavigationItem: React.FC<Props> = ({ name, to, children }) => {
...
   &:hover {
     text-decoration: none;
   }
 `;
 
 const NavigationList = styled.ul`
-  width: 100%;
-  margin-left: 30px;
-  padding: 0;
+  padding-left: 12px;
```
KaWaite, line 40: metrics?

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: yes, I feel when I develop, if I use a flex component which is a little hard to use, the properties is different from native css, and because of using flex-component , I made some mistakes. Here of course I can still use flex-component, so you want to use back the flex-component ?

src/components/molecules/Settings/Navigation/index.tsx
lavalse: ok. sure

src/components/molecules/Settings/Project/EditableItem/index.tsx
```diff
@@ -130,7 +130,7 @@ const EditableItem: React.FC<Props> = ({
 
 const StyledIcon = styled(Icon)`
   padding: 0;
-  margin-left: 15px;
+  margin-left: 16px;
```
KaWaite, line 133: metrics?

src/components/molecules/Settings/ProjectList/ProjectCell/index.tsx
```diff
@@ -29,47 +30,74 @@ export type Props = {
...
-const Wrapper = styled.div<{ project: Project }>`
+const StyledWrapper = styled.div<{ project: Project }>`
   background: ${props =>
-    props.project.imageUrl ? `url(${props.project.imageUrl})` : props.theme.colors.bg[4]};
+    props.project.imageUrl ? `url(${props.project.imageUrl})` : `url(${defaultProjectImage})`};
   background-size: cover;
-  color: ${props => props.theme.projectCell.text};
-  padding: 10px;
+  background-position: center;
+  ${props => !props.project.imageUrl && "background-size: 400px 240px;"};
```
KaWaite, line 66: Move this up to and combine with the `background-size: cover;`

src/components/molecules/Settings/NavigationItem/index.tsx
lavalse: yes, but we really really need metrics to use in all padding and marging? like, we will not change any space when we switch the theme...  any way, it's really Crush me about using metrics 💀

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: ok, I wil fix it

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
KaWaite: If you find it hard to use, the team can talk more about its usefulness. For now though we have been trying to go in the direction of using it when appropriate. If it is really hard to use perhaps discuss with @HideBa when we have time.

lavalse: and also the flex-component don't have align-content property . align-items and align-content are different ... I guess

lavalse: but I remembr @HideBa is agree with us can not use flex-component , right? Or we must ? @HideBa @KaWaite ?

src/components/molecules/Settings/ProjectList/ProjectList/index.tsx
```diff
@@ -74,39 +70,39 @@ const ProjectList: React.FC<Props> = ({
...
       {loading ? (
         <div>
           <Loading />
         </div>
       ) : (
         <>
-          <Divider />
           {open && (
-            <Flex wrap="wrap">
+            <ProjectListCotainner>
```
KaWaite, line 82: Typo: ProjectListContainer

src/components/molecules/Settings/Section/index.tsx
```diff
@@ -15,47 +14,35 @@ const Section: React.FC<Props> = ({ title, actions, children }) => {
...
-      </SectionItem>
+      <SectionItem>{children}</SectionItem>
     </Wrapper>
   );
 };
 
-const Wrapper = styled.div`
-  max-width: 100%;
-`;
+const Wrapper = styled.div``;
```
KaWaite, line 30: This doesn't need to be a styled component

src/components/molecules/Settings/SettingPage/index.tsx
```diff
@@ -35,36 +40,104 @@ const SettingPage: React.FC<Props> = ({
...
+  margin-left: 264px;
+  padding: 32px 64px;
   box-sizing: border-box;
+
+  @media only screen and (max-width: 1024px) {
+    margin: 0 auto;
+  }
+`;
+
+const Wrapper5 = styled.div`
```
KaWaite, line 108: Any reason we have Wrapper, Wrapper3 and Wrapper5? Can you name them more clearly?

src/components/molecules/Settings/SettingsHeader/index.tsx
```diff
@@ -20,12 +20,12 @@ const SettingsHeader: React.FC<Props> = ({ title, currentWorkspace, currentProje
...
   return (
     <Wrapper>
-      <Text size="xl" color={theme.main.strongText} weight="bold">
+      <Text size="xl" color={theme.main.strongText} weight="normal">
         {workspace} {workspace && (title || currentProject) && " / "}
         {currentProject} {title && currentProject && " / "}
         {title}
       </Text>
-      <Text size="m" otherProperties={{ marginTop: "12px" }}>
+      <Text size="m" className={theme.main.text} otherProperties={{ marginTop: "12px" }}>
```
KaWaite, line 28: Did you mean this className to be color??

src/components/molecules/Settings/WorkspaceList/WorkspaceCell/index.tsx
```diff
@@ -22,52 +22,73 @@ const WorkspaceCell: React.FC<Props> = ({ className, team, personal, onSelect })
...
-  color: ${props => props.theme.colors.text.strong};
-  padding: 10px;
+  background: ${props => props.theme.colors.bg[3]};
+  box-sizing: border-box;
   box-shadow: 0 0 5px ${props => props.theme.projectCell.shadow};
-  margin: 10px 0;
-  padding: 28px;
-  width: 100%;
-  height: 223px;
+  padding: 16px 24px;
```
KaWaite, line 61: metrics?

src/components/molecules/Settings/WorkspaceList/WorkspaceCell/index.tsx
```diff
@@ -22,52 +22,73 @@ const WorkspaceCell: React.FC<Props> = ({ className, team, personal, onSelect })
...
+  justify-content: space-between;
   cursor: pointer;
+
+  &:hover {
+    background: ${({ theme }) => theme.colors.bg[4]};
+  }
+`;
+
+const TopWrapper = styled.div`
+  display: flex;
```
KaWaite, line 74: Flex component?

src/components/molecules/Settings/WorkspaceList/WorkspaceCell/index.tsx
```diff
@@ -22,52 +22,73 @@ const WorkspaceCell: React.FC<Props> = ({ className, team, personal, onSelect })
...
+  justify-content: space-between;
+  align-items: center;
+`;
+
+const BottomWrapper = styled.div`
```
KaWaite, line 79: Flex component?

src/components/organisms/Settings/WorkspaceList/index.tsx
```diff
@@ -29,7 +29,7 @@ const WorkspaceList: React.FC<Props> = ({ teamId }) => {
 
   return (
     <SettingPage teamId={teamId}>
-      <SettingsHeader title={intl.formatMessage({ defaultMessage: "Workspaces list" })} />
+      <SettingsHeader title={intl.formatMessage({ defaultMessage: "My Workspaces list" })} />
```
KaWaite, line 32: My Workspace List
Above we have "My Account". Either both should have only the first word capitalized or all capitalized. Stylistically it is up to you

KaWaite (changes requested):
Please look at my comments and make sure @rot1024 adds the Japanese translations

src/components/molecules/Settings/NavigationItem/index.tsx
KaWaite: The idea behind the metrics (like colors, etc) is that we can change them in one place (the Theme files) and they'll change everywhere. Metrics is a little debatable though as changes to metrics will probably be accompanied by larger changes, whereas color changes are relatively minor and changing one source is really convenient.
But anyways, as much as possible we are trying to consistently make use of the available metrics.

src/components/molecules/Common/AssetModal/AssetListItem/index.tsx
KaWaite: Since it seems we are keeping it this way, double and triple check you've deleted all reference to isImage

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: and here, because I need to get the width of div, it's just difficult using Flex-component here

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: oh, yes, it's ture, thanks.

src/components/molecules/Settings/Field/index.tsx
lavalse: and here, because the Flex-component don't have justify-items property, I think it's better not to use Flex-component ..

src/components/molecules/Settings/Logo/index.tsx
lavalse: ok, I just think it will be easy. I will remove the component

src/components/molecules/Settings/NavigationItem/index.tsx
lavalse: ok, thanks

src/components/molecules/Settings/Project/EditableItem/index.tsx
lavalse: thanks

src/components/molecules/Settings/ProjectList/ProjectList/index.tsx
lavalse: I'm blind , thanks

src/components/molecules/Settings/Section/index.tsx
lavalse: ok, thanks

src/components/molecules/Settings/SettingPage/index.tsx
lavalse: yes, I these wrappers are for structures, I name them more clearly, any good ideas? @KaWaite

src/components/molecules/Settings/SettingsHeader/index.tsx
lavalse: oh, yes, thanks

src/components/molecules/Settings/WorkspaceList/WorkspaceCell/index.tsx
lavalse: how about future, because align-iteams here

src/components/organisms/Settings/WorkspaceList/index.tsx
lavalse: yesyes

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
lavalse: thanks, I will use another way.

src/components/molecules/Common/AssetModal/VirtualAssetCard/index.tsx
lavalse: thanks, I will switch to css grid

src/components/organisms/Settings/ProjectList/hooks.ts
lavalse: on line: 100 , it refetched, so I don't need refetch here.. @rot1024 
      
const scene = await createScene({
        variables: { projectId: project.data.createProject.project.id },
        refetchQueries: ["Me"],
      });

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -57,16 +59,17 @@ const Wrapper = styled(Flex)<{ selected?: boolean; cardSize?: CardSize }>`
   box-shadow: 0 6px 6px -6px ${props => props.theme.colors.other.black};
   border: 1px solid
     ${props => (props.selected ? `${props.theme.assetCard.highlight}` : "transparent")};
-  margin: ${({ cardSize }) => (cardSize === "small" ? "5px" : "10px")};
+  /* margin: 12px; */
```
KaWaite, line 62: Don't forget commented out code

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -23,22 +23,26 @@ export type Props = {
   assets?: Asset[];
   isMultipleSelectable?: boolean;
   accept?: string;
-  onCreateAsset?: (file: File) => void;
+  onCreateAsset?: (files: FileList) => void;
+  onRemove?: (assets: string[]) => void;
```
KaWaite, line 27: Sorry I should have suggested earlier, but just to make it very clear naming the passed value assetIds would be best I think. We aren't passing assets so it isn't quite correct.

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -166,35 +184,44 @@ const AssetContainer: React.FC<Props> = ({
...
+  min-height: 400px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
 `;
 
-const AssetList = styled(Flex)<{ layoutType?: LayoutTypes }>`
-  ${({ layoutType }) => layoutType === "list" && "flex-direction: column;"}
-  max-height: 458px;
+const ButtonWrapper = styled(Flex)``;
```
KaWaite, line 198: This styled component isn't necessary. Just use Flex directly above.

src/components/molecules/Common/Header/index.tsx
```diff
@@ -107,13 +107,17 @@ const Wrapper = styled.header`
   width: 100%;
   height: ${metrics.headerHeight}px;
   background: ${props => props.theme.header.bg};
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
```
KaWaite, line 112: Little confused why this is necessary. The Content component declared below has flex. Also this Wrapper only has Content as its child so flex and flex-direction as far as I can tell would have no affect. Let me know if I am missing something.

src/components/molecules/Settings/Account/AccountSection/index.tsx
```diff
@@ -2,10 +2,10 @@ import React, { useEffect, useState } from "react";
 import PasswordModal from "@reearth/components/molecules/Settings/Account/PasswordModal";
 import Section from "@reearth/components/molecules/Settings/Section";
 import EditableItem from "@reearth/components/molecules/Settings/Project/EditableItem";
+import Field from "@reearth/components/molecules/Settings/Field";
 import Icon from "@reearth/components/atoms/Icon";
-import { styled, useTheme } from "@reearth/theme";
+import { styled, colors } from "@reearth/theme";
```
KaWaite, line 7: Sorry after our discussion about importing colors like this, @issmail-basel pushed his theme branch and I saw he was removing all directly used colors in this way.
So rather than importing like this and then using like `colors.text.strong` you should be able to use colors through the theme like `${({theme}) => theme.XX.XXXX};`
That was my bad, my apologies

src/components/molecules/Settings/NavigationItem/index.tsx
```diff
@@ -1,6 +1,8 @@
 import React from "react";
 import { styled, fonts } from "@reearth/theme";
 import { Link } from "@reach/router";
+import { colors } from "@reearth/theme";
```
KaWaite, line 4: Again sorry looks like we won't be importing the colors directly when we have multiple themes. Same as above comment, use colors directly from theme

src/components/molecules/Settings/ProjectList/ProjectCell/index.tsx
```diff
@@ -29,47 +31,77 @@ export type Props = {
...
   box-shadow: 0 0 5px ${props => props.theme.projectCell.shadow};
-  margin: 10px 0;
-  padding: 28px;
-  // position: relative;
-  width: 100%;
-  height: 223px;
+  height: 240px;
+`;
+
+const Wrapper = styled.div<{ project: Project; isHover?: boolean }>`
```
KaWaite, line 71: The project prop isn't being used. It is used in the StyledWrapper above which it looks like you split into two styled components. Remove here and above in the return

src/components/molecules/Settings/Section/index.tsx
```diff
@@ -11,51 +11,37 @@ export type Props = {
 const Section: React.FC<Props> = ({ title, actions, children }) => {
   const theme = useTheme();
   return (
-    <Wrapper>
+    <>
```
KaWaite, line 14: Why can't you use a div here? I am not 100% but to me this is kinda bad for HTML semantics. I know divs don't have meaning(compared to other HTML elements like header, footer, etc) but I don't see a reason not to use a div to clump this related stuff together.
I try to only use React fragments when I absolutely have to or it makes sense (ie. I have a Modal component that will render conditionally outside of the main content of that component).

KaWaite (approved):
GTG

translations/en.yml
```diff
@@ -327,45 +354,22 @@ src:
...
             '2999985900': Workspace name
         WorkspaceList:
           WorkspaceCell:
-            '293600023': 'Team Members:'
-            '2968463849': This is your personal account
-            '3038786653': No Title Project
+            '808766300': No Title Workspace
+            '974692228': >-
+              This is your personal workspace, your projects and resource will
+              be managed in this workspace.
```
KaWaite, line 371: This is your personal workspace. Your projects and resources will be managed in this workspace.

translations/ja.yml
```diff
@@ -352,19 +354,19 @@ src:
           '3590336714': エクスポート
       Settings:
         Account:
-          '3198092477': アカウント
+          '3198092477': 'アカウント'
```
KaWaite, line 357: Not sure the ' ' will show up in app, but just to be sure and to keep consistent let's remove them ;)

KaWaite (approved):
So just a couple comments but looks good to me (again :P). We are starting to get a lot of untranslated parts so make sure @rot1024 or @HideBa update the Japanese.

src/components/organisms/Settings/Workspace/Asset/hooks.ts
```diff
@@ -8,6 +8,15 @@ import {
 } from "@reearth/gql";
 import { useLocalState } from "@reearth/state";
 
+export type Asset = {
```
HideBa, line 11: Is this type of definition used? If not, please delete it. 👍

translations/ja.yml
```diff
@@ -27,6 +27,7 @@ src:
             '2854191800': ''
             '2906553848': ''
             '3057159632': 画像アップロード
+            '3645414921': ''
```
HideBa, line 30:
```suggestion
            '3645414921': 削除
```

translations/ja.yml
```diff
@@ -216,16 +217,21 @@ src:
...
         Project:
           ArchivedMessage:
             '1354606874': 通知
             '4112013314': >-
               プロジェクトがアーカイブされました。アーカイブ済みプロジェクトの設定を変更する場合は、プロジェクトのアーカイブを解除してから変更可能です
+          BasicAuthSection:
+            '256145410': ''
+            '2237029987': ''
+            '2663586870': ''
+            '3076470241': ''
```
HideBa, line 234:
```suggestion
            '256145410': ベーシック認証を有効化
            '2237029987': パスワード
            '2663586870': ユーザー名
            '3076470241': ベーシック認証
```

Thanks for helping my fault. 🙇

translations/ja.yml
```diff
@@ -251,9 +257,34 @@ src:
...
+            PluginSection:
+              '3060204202': ''
+              PluginInstall:
+                '149186996': ''
+                '311550959': ''
+                '1361263844': ''
+                '1889877734': ''
+                '3129492077': ''
+              PluginList:
+                '1933254449': ''
```
HideBa, line 284:
```suggestion
              '2065124393': このデータセットを削除
              '3112669357': 本当に削除してよろしいですか？
              '3528672691': キャンセル
              '3645414921': 削除
            DatasetSection:
              '293739158': データセットを追加
              '3833270271': データセット
          Plugin:
            PluginAccordion:
              PluginAccordionItem:
                '31593394': アンインストール
                '730907263': インストール済み
                '3480550112': インストール
            PluginSection:
              '3060204202': 現在開発中、もうしばらくお待ち下さい。
              PluginInstall:
                '149186996': アップロード済みプラグイン
                '311550959': GitHubプライベートレポジトリ
                '1361263844': GitHubパブリックレポジトリ
                '1889877734': プラグイン一覧
                '3129492077': PCからZipファイルをアップロード
              PluginList:
                '1933254449': 開発者向けオプション
```

Thanks

translations/ja.yml
```diff
@@ -251,9 +257,34 @@ src:
...
           ProfileSection:
-            '583832611': プロジェクト名
-            '1534468327': プロフィール
+            '636461959': ''
+            '1513791786': ''
```
HideBa, line 287:
```suggestion
            '636461959': 名前
            '1513791786': プロジェクト情報
```

HideBa (approved):
Thanks for your awesome large changes. I've left some comments. Please check them. 👍

src/components/organisms/Settings/Workspace/Asset/hooks.ts
lavalse: thanks

translations/ja.yml
lavalse: sure

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -35,8 +33,12 @@ const AssetCard: React.FC<Props> = ({
       selected={selected}
       cardSize={cardSize}
       onClick={() => onCheck?.(!check)}>
-      <ImgWrapper cardSize={cardSize} url={url}>
-        {!isImage && <Icon icon="file" />}
+      <ImgWrapper cardSize={cardSize}>
+        {/\.(jpg|jpeg|png|gif|svg|webp|GIF|JPG|JPEG|PNG|SVG|WEBP)$/.test(url) ? (
+          <PreviewImage url={url} />
+        ) : (
+          <Icon icon="file"></Icon>
```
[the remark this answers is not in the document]
lavalse: fixed

rot1024 (approved):

translations/ja.yml
```diff
@@ -149,7 +150,7 @@ src:
               PropertyLinkPanel:
                 '773082751': 設定をリセット
                 '910188724': 選択可能なアイテムがありません
-                '1055097961': 他のデータセットにリンク
+                '1055097961': ''
```
KaWaite, line 153: Missing translation

translations/ja.yml
```diff
@@ -293,47 +320,22 @@ src:
...
           ProfileSection:
             '2999985900': ワークスペース名
         WorkspaceList:
           WorkspaceCell:
-            '293600023': チームメンバー
-            '2968463849': パーソナルアカウント
-            '3038786653': プロジェクト名未設定
+            '808766300': ''
+            '974692228': ''
+            '3230245813': ''
```
KaWaite, line 336: Missing translations

KaWaite (approved):
A couple translations are missing (I left comments where they are) but other than that looks good!

PR #21, opened 2021-06-23
fix: menu button width

issmail-basel:
# Overview
When the menu has been opened the style should be kept well

## What I've done
I made the menu min-width of 200px
but if the button is wider the menu will take the whole width (as shown in the screenshots)


HideBa (approved):
As long as @lavalse approves how it looks, looks good to me!

lavalse (approved):
Seems good for me

PR #22, opened 2021-06-24
feat: plugins settings page, install/uninstall plugins

HideBa:
# Overview
Show installable plugins list on the plugin page

## What I've done
* Impl query to get installable plugins list
* Show if the plugin has been installed or not by fetching installed plugins

## What I haven't done
* Mutation to install the plugin
* Write test code for this component because I got an error which attributes Jest. I thought I shouldn't consume my time for that now.

## How I tested
* Opened the plugin page and checked `Re:Earth	official plugin` is displayed since it has been installed
* `Re:Earth official plugin` shows that it has been installed

## Screenshot

スクリーンショット 2021-06-24 11 27 56

* The logic to show if the plugins have been installed or not
*


src/components/molecules/Settings/Workspace/Plugin/PluginSection/index.tsx
```diff
@@ -1,15 +1,19 @@
-import React from "react";
+import React, { useState } from "react";
 
-import { styled, useTheme } from "@reearth/theme";
+// import { styled, useTheme } from "@reearth/theme";
```
HideBa, line 3: These commented parts will be uncommented unless the full plugin feature is released.

src/components/organisms/Settings/Project/Plugin/hooks.ts
```diff
@@ -1,28 +1,37 @@
 import { useLocalState } from "@reearth/state";
+import { useInstallablePluginsQuery, useInstalledPluginsQuery } from "@reearth/gql";
+import { useMemo } from "react";
```
KaWaite, line 3: All react imports should be at the very top

KaWaite (dismissed):
One minor change requested so after that is changed  looks good!

src/components/organisms/Settings/Project/Plugin/hooks.ts
HideBa: Thank you!

src/components/molecules/Settings/Workspace/Plugin/PluginAccordion/index.test.tsx
```diff
@@ -5,18 +5,18 @@ import PluginAccordion from "./index";
...
-      id: "fugafuga",
-      thumbnail: `${process.env.PUBLIC_URL}/sample.svg`,
+      thumbnailUrl: `${process.env.PUBLIC_URL}/sample.svg`,
       title: "Sample2",
       isInstalled: false,
       bodyMarkdown: "# Fuga ## Fuga",
+      author: "reearth",
     },
   ];
   test("Accordion should be rendered", () => {
```
issmail-basel, line 22: Please just out of curiosity:
1- What type of tests is this? `import { render, screen } from "@reearth/test/utils";` is only used twice in all Re:Earth project.
2- If it is using `getByText`, I guess it's checking the UI (E2E) so why not using integrated Cypress library to test?

src/components/molecules/Settings/Workspace/Plugin/PluginAccordion/index.tsx
```diff
@@ -19,17 +19,17 @@ export type PluginAccordionProps = {
 
 const PluginAccordion: React.FC<PluginAccordionProps> = ({ className, items }) => {
   const theme = useTheme();
-  return (
+  return items ? (
     <Accordion
       className={className}
       allowMultipleExpanded
       itemBgColor={theme.colors.bg[3]}
       items={items?.map(i => {
```
issmail-basel, line 27: Just a personal preference to use clear name of variables (even inside iterations), so it would be:
`items={items?.map(item => {` instead of using `i`
What do you think?

src/components/molecules/Settings/Workspace/Plugin/PluginAccordion/index.tsx
```diff
@@ -19,17 +19,17 @@ export type PluginAccordionProps = {
 
 const PluginAccordion: React.FC<PluginAccordionProps> = ({ className, items }) => {
   const theme = useTheme();
-  return (
+  return items ? (
```
issmail-basel, line 22: just a suggestion of using
`items &&`
instead of ternary operator

src/components/organisms/Settings/Project/Plugin/hooks.ts
```diff
@@ -1,28 +1,38 @@
...
+          bodyMarkdown: p.description,
+          author: p.author,
+          thumbnailUrl: p.thumbnailUrl,
+          isInstalled: !!installedPluginIds?.includes(p.name), //TODO: After back-end decide how to generate plugin's id, fix here.
+        }))
+      : [];
+  }, [installedPluginIds, rawPluginsData]);
 
-  return { currentTeam, currentProject, plugins: samplePlugins };
+  const loading = sceneLoading || pluginLoading;
```
issmail-basel, line 36: I'm not sure, but I want to ask.
Does this line works correctly?
Loading is a constant and assigned once, what's the purpose of it?

issmail-basel (commented):
Code looks fine
But, please check my comments, maybe there is something that could help.
Then I'll approve

src/components/molecules/Settings/Workspace/Plugin/PluginAccordion/index.test.tsx
HideBa: 1. That is a component test for react component. We use `react-testing-library` and can test if the components are rendered as expected.  Our project uses some special like loading i18n from YAML files, receive style from theme providers, etc. So we need to mock that kind of thing and that's why I defined the custom `render` function. Also, it can receive the mock of GQL data and can be used for components that have dependencies on GQL.

HideBa: 2. Cypress is used for E2E test. This is the component test like a unit test for components.

src/components/molecules/Settings/Workspace/Plugin/PluginAccordion/index.tsx
HideBa: I assume others can guess what it is but the clearer the better. I choose your idea. 👍

src/components/molecules/Settings/Workspace/Plugin/PluginAccordion/index.tsx
HideBa: The reason why I used the ternary operator is to tell it returns `null` explicitly. Kind of developers preference? haha

src/components/organisms/Settings/Project/Plugin/hooks.ts
HideBa: It's a const value, but the component will be rendered when the loading state returned by Apollo Query has been changed. So the `loading` will be also updated right? @rot1024

rot1024: Yes, no problem. The component function will be called every render time, so all lines are also executed every time. That is the same as normal functions of JavaScript. Only the value returned by the hooks may change each time the component is rendered.

src/test/utils.tsx
```diff
@@ -3,30 +3,12 @@ import { render as rtlRender } from "@testing-library/react";
...
-      },
-    },
-  },
-];
-const render = (ui: React.ReactElement, { ...renderOptions } = {}) => {
+const render = (
+  ui: React.ReactElement,
+  queryMocks?: readonly MockedResponse<Record<string, any>>[],
+  { ...renderOptions } = {},
+) => {
```
rot1024, line 11: On an unrelated topic, I don't like this helper function for unit testing. It's better to directly import testing-library in each test case, so that we can see at a glance what context the component requires just by looking at the test code.

src/test/utils.tsx
```diff
@@ -3,30 +3,12 @@ import { render as rtlRender } from "@testing-library/react";
...
   const Wrapper: React.FC<{ children?: React.ReactNode }> = ({ children }) => {
     return (
       <ThemeProvider>
```
HideBa, line 14: @rot1024 
>On an unrelated topic, I don't like this helper function for unit testing. It's better to directly import testing-library in each test case so that we can see at a glance what context the component requires just by looking at the test code.

The reason why I made this helper function is to test the components which have something to do with `theme` and i18n we need to extend the testing library's `render` function. If we need this helper function, I thought it's more convenient to add gqlMock as well. Is it better to do this on each component test?

rot1024: Basically, I think so. Test code can be thought of as a description of a specification. And the fact that the component depends on some context is also a good specification.

It is not easy to read test code where such specifications cannot be read from the test code or cannot be understood without jumping to the import destination.

Also, when multiple providers are provided in a helper function, it is difficult to know which context the component depends on without reading the component implementation.

HideBa: Thanks. I got the idea. As a refactoring task, we'll fix this part. For now, let it pass this review.

HideBa:
@rot1024 If everything is fine, plz approve.

rot1024 (approved):
CI should be fixed

HideBa:
@rot1024 @KaWaite @Basel-Issmail @lavalse 

I've updated this branch since the requirements had been changed.

## What I did
* Show installed plugins list
* Install plugins by uploading zip files from local
* Install plugins from GitHub public repo
* Uninstall selected plugin
* Added Confirmation modal which can be used to confirm the action anywhere on Re:Earth
* Added Japanese text on translation file

lavalse:
image
Sorry @HideBa , but seems wired line comes out in the uninstall modal. Can we remove it, just like other modals?
image

HideBa:
@lavalse 

>but seems wired line comes out in the uninstall modal. Can we remove it, just like other modals?

OK! Thank you for letting me know that!

rot1024:
Re:Earth official plugin cannot be deleted. It's OK in UI?

lavalse:
@rot1024 For right design, we have a page which show all the official plugins like this 
image
Users can install and uninstall , but the list will always show . and in Coustom plugin page 
image
we can upload ( import ) and delete.

Now we don't have that structure.
In the design, I think they are different between install / uninstall / delete
For offical plugin user can only install and uninstall
For cunstom plugin , user can upload , then install / uninstall , and of cause delete

so ... @HideBa you should think a little bit, or ask me to change design

rot1024:
I mean "Re:Earth official plugin" is the built-in plugin. It cannot be deleted.

lavalse:
image
Very Sorry. I found here , the icon didn't useTheme, so if you can update here also help muh. 🙏  @HideBa

rot1024:
If I press the ESC key, it will take me to another screen:

recording


<details><summary>Browser metadata</summary>





```
Path:      /settings/workspace/01f6h7cs13wjx8cj6r2dpmfeav
Browser:   Chrome 92.0.4515.107 on Mac OS 10.15.7
Viewport:  1298 x 764 @1x
Language:  ja
Cookies:   Enabled
```

[Open in BrowserStack](https://www.browserstack.com/user/try-live?url=https%3A%2F%2Fdeploy-preview-22--reearth-web.netlify.app%2Fsettings%2Fworkspace%2F01f6h7cs13wjx8cj6r2dpmfeav&os=OS%20X&os_version=Catalina&browser=Chrome&browser_version=92.0&resolution=1298x764&speed=1&start=true&ref=netlify-source)



</details>

[Open Deploy Preview](https://deploy-preview-22--reearth-web.netlify.app/settings/workspace/01f6h7cs13wjx8cj6r2dpmfeav) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=610b8c89bd07ab0007a6a1e0&commentID=610b90e17c3b1e8d7a80573b&resolution=resolved)

rot1024:
If I press the ESC key, it will take me to another screen:

Kapture 2021-08-05 at 16 20 35

HideBa:
@lavalse 
>Very Sorry. I found here , the icon didn't useTheme, so if you can update here also help muh. 🙏 @HideBa

Sorry for that! I've just fixed, plz check again.

rot1024 (dismissed):

src/components/molecules/Settings/Project/Plugin/PluginAccordion/index.test.tsx
```diff
@@ -1,29 +1,36 @@
...
+  // ];
+  // test("Accordion should be rendered", () => {
+  //   render(<PluginAccordion items={items} uninstallPlugin={mockUninstallPlugin} />);
+  // });
+  // test("PluginAccordion should display header and not display body", () => {
+  //   render(<PluginAccordion items={items} uninstallPlugin={mockUninstallPlugin} />);
+  //   expect(screen.getByText("Sample")).toBeInTheDocument();
+  // });
+  test("dummy", () => {
+    console.log("dummy test---");
```
issmail-basel, line 33: Just a suggestion
Maybe it is better to not use console.log even if it's in a test file.

issmail-basel (approved):
Everything seems good

src/components/molecules/Settings/Project/Plugin/PluginAccordion/index.test.tsx
HideBa: I'll fix this test after the event 👍

HideBa:
@lavalse Can I merge this?

rot1024 (approved):

PR #25, opened 2021-07-08
feat: add  delete assets confirm modal and fix bugs

lavalse:
# Overview
image

## What I've done
- add a delete assets modal in assets setting page
- fix project page no data bug
- fix no create new asset bug

## Thanks
@Basel-Issmail @KaWaite

### DIFF
--- a/src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
+++ b/src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
@@ -1,11 +1,60 @@
 import React from "react";
+import { useIntl } from "react-intl";
+
+import Text from "@reearth/components/atoms/Text";
 import Modal from "@reearth/components/atoms/Modal";
+import Divider from "@reearth/components/atoms/Divider";
+import Button from "@reearth/components/atoms/Button";
+
+import { styled } from "@reearth/theme";
+import { metricsSizes } from "@reearth/theme/metrics";
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface Props {}
+export interface Props {
+  isVisible: boolean;
+  onClose?: () => void;
+  handleRemove?: () => void;
+}
 
-const AssetDeleteModal: React.FC<Props> = () => {
-  return <Modal title="assets delete ?" isVisible={true} size="sm"></Modal>;
+const AssetDeleteModal: React.FC<Props> = ({ isVisible, onClose, handleRemove }) => {
+  const intl = useIntl();
+  return (
+    <Modal
+      title="Delete assets"
+      isVisible={isVisible}
+      size="sm"
+      onClose={onClose}
+      button1={
+        <Button
+          text={intl.formatMessage({ defaultMessage: "Delete" })}
+          buttonType="danger"
+          onClick={handleRemove}
+        />
+      }
+      button2={
+        <Button
+          text={intl.formatMessage({ defaultMessage: "Cancel" })}
+          buttonType="secondary"
+          onClick={onClose}
+        />
+      }>
+      <Divider margin="0" />
+      <Message size="m">
+        {intl.formatMessage({
+          defaultMessage:
+            "You are about to delete assets from the workspace. Any place which use this asset will cause a error. And this is an action that cannot be undone.",
+        })}
+      </Message>
+      <Message size="m">
+        {intl.formatMessage({
+          defaultMessage: "Are you sure you would like to delete these assets?",
+        })}
+      </Message>
+    </Modal>
+  );
 };
 
+const Message = styled(Text)`
+  margin-top: ${`${metricsSizes["2xl"]}px`};
+`;
 export default AssetDeleteModal;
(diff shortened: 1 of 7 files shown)


src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
```
KaWaite, line 1: Can we move this component into the AssetModal folder in Common? Since it is used in the AssetContainer and I can't see it being used elsewhere, it would make more sense to keep it together with the AssetModal

src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+          text={intl.formatMessage({ defaultMessage: "Cancel" })}
+          buttonType="secondary"
+          onClick={onClose}
+        />
+      }>
+      <Divider margin="0" />
+      <Message size="m">
+        {intl.formatMessage({
+          defaultMessage:
+            "You are about to delete assets from the workspace. Any place which use this asset will cause a error. And this is an action that cannot be undone.",
```
KaWaite, line 45:
```suggestion
            "You are about to delete one or more assets from the workspace. Please make sure no selected assets are in use or else an error will occur. This cannot be undone!",
```
Is it true that an error will occur? Or just no image will be shown???

src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+        })}
+      </Message>
+      <Message size="m">
+        {intl.formatMessage({
+          defaultMessage: "Are you sure you would like to delete these assets?",
```
KaWaite, line 50: Maybe the above message is enough? If you think this should be here, maybe "Are you sure you would like to delete the selected asset(s)?"

src/components/organisms/Dashboard/hooks.ts
```diff
@@ -145,7 +145,9 @@ export default (teamId?: string) => {
       .filter((project): project is Project => !!project);
   }, [team?.projects.nodes]);
 
-  const [createNewProject] = useCreateProjectMutation();
+  const [createNewProject] = useCreateProjectMutation({
+    refetchQueries: ["Project"],
```
KaWaite, line 149: You have added this refetchQueries but in the createProject function we are still using refetch() as well.
I think we only need one. Also, if you are adding refetchQueries that probably means refetch() wasn't working properly, so either way it should be removed from the callback!

src/components/organisms/Settings/ProjectList/hooks.ts
```diff
@@ -23,12 +25,15 @@ const toPublishmentStatus = (s: PublishmentStatus) =>
...
   const navigate = useNavigate();
+  const intl = useIntl();
 
   const [modalShown, setModalShown] = useState(false);
   const openModal = useCallback(() => setModalShown(true), []);
 
   const { data, loading, refetch } = useMeQuery();
-  const [createNewProject] = useCreateProjectMutation();
+  const [createNewProject] = useCreateProjectMutation({
+    refetchQueries: ["Project"],
```
KaWaite, line 35: If we are doing a refetch() in the createProject function for MEQuery do we need the refetchQuery attached here? 
Pretty sure we just need one (and the refetch() makes sense since we are fetching AFTER the project is created AND added to the scene).
Please let me know if I am mistaking the logic!

src/components/organisms/Settings/ProjectList/hooks.ts
```diff
@@ -125,6 +131,21 @@ export default () => {
   });
   const assets = assetsData?.assets.nodes.filter(Boolean) as AssetNodes;
 
+  const [createAssetMutation] = useCreateAssetMutation();
```
KaWaite, line 134: Probably a standard we need to discuss and start using across the app, but for now in this file it seems all the useXXXXMutation() declarations are put together above, so if you could move this up (or move the above declarations to just above where they are used(like how you wrote here))

KaWaite (commented):
Please check the couple comments!

src/components/organisms/Settings/ProjectList/hooks.ts
lavalse: Like yesterday I pairprogram with Basel , we found seems the MEQuery didn't update the projects ... that's why here to refetch the projects. I hope we can have a discuss @KaWaite @Basel-Issmail for this problem

src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
lavalse: Yes, I hope we can keep this I will fix

src/components/molecules/Settings/Workspace/Asset/AssetDeleteModal/index.tsx
lavalse: just the image will not show. Any good Suggestions?

KaWaite (approved):

issmail-basel (approved):
Seems good

KaWaite (approved):

PR #26, opened 2021-07-13
fix: visualizer bugs

rot1024:
Fixes:

- Layers without title are not recognized by storytelling widget
- Menu widget is not displayed
- Camera won't fly with splash screen widget
- Timing of the start of splash screen is too early
- Disable Cesium's default double click event
- Marker label position is opposite to the actual display


HideBa (approved):
LGTM

PR #29, opened 2021-07-14
fix: menu widget bugs

KaWaite:
# Overview
Menu widget should show icon when its type is set to icon or texticon, but it didn't show.

## What I've done
- Fixed Icon component as its logic wasn't right. Now it works with non-reearth local urls.
- Fixed issue where you couldn't close a dropdown from menu type

## What I haven't done
- Had to remove the ScreenSpaceEventHandler from the widget since it wasn't working. This only effects clicking away on the cesium area. Clicking on other UI will still hide the dropdown
- Still a bug where all buttons show the same menu items

## How I tested
Added an image and tested different button types to make sure each worked correctly.

## Screenshot
Screen Shot 2021-07-14 at 15 47 17


src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
```diff
@@ -66,26 +60,53 @@ export default function ({
...
+        }
+      } else {
+        let link = "buttonLink" in b ? b.buttonLink : "menuLink" in b ? b.menuLink : undefined;
+        const splitLink = link?.split("/");
+        if (splitLink?.[0] !== "http:" && splitLink?.[0] !== "https:") {
+          link = "https://" + link;
+        }
+        window.open(link, "_blank", "noopener");
+      }
+      setVisibleMenuButton(undefined);
```
rot1024, line 92:
```suggestion
      setVisibleMenuButton(undefined);
      const t = "buttonType" in b ? b.buttonType : "menuType" in b ? b.menuType : undefined;
      if (t === "menu") {
        setVisibleMenuButton(v => (v === b.id ? undefined : b.id));
        return;
      } else if (t === "camera") {
        const camera =
          "buttonCamera" in b ? b.buttonCamera : "menuCamera" in b ? b.menuCamera : undefined;
        if (camera) {
          flyTo?.(camera, { duration: 2000 });
        }
      } else {
        let link = "buttonLink" in b ? b.buttonLink : "menuLink" in b ? b.menuLink : undefined;
        if (!link) return;
        const splitLink = link?.split("/");
        if (splitLink?.[0] !== "http:" && splitLink?.[0] !== "https:") {
          link = "https://" + link;
        }
        window.open(link, "_blank", "noopener");
      }
```

rot1024 (changes requested):

PR #30, opened 2021-07-14
fix: selection bugs

rot1024:
Fixes:

- Some errors from React in console
- Widget and scene cannot be selected when layer is selected
- Block cannot be selected when other block is selected
- Text block is not editable by double clicking
- Move widgets inside engines
- Format


KaWaite (approved):
Looks like we both made some changes to the Menu widget. Anyways, except for the error(maybe a missed isSelect) looks good!

PR #40, opened 2021-07-16
feat: support multi-line infobox titles

rot1024:
# Overview

## What I've done

image image

Support multi-line infobox titles. Long word will also break automatically.


KaWaite (approved):

HideBa (approved):

HideBa:
@rot1024 Is is adopted on published page as well?

rot1024:
I made changes in InfoBoxFrame and it is used by both InfoBox and SimpleInfoBox so it should work in published pages, but not testable. (I will work on it next)

PR #45, opened 2021-07-16
fix: show layers in storytelling without names

KaWaite:
# Overview

## What I've done
Just a bandaid fix as a better fix is coming with the reworked plugin system!


src/components/molecules/Common/plugin/builtin/widgets/storytelling.tsx
```diff
@@ -56,13 +56,10 @@ const Storytelling: WidgetComponent<Property, PluginProperty> = ({ property }) =
...
-          if (title) {
-            return {
-              ...story,
-              title,
-            };
-          }
-          return undefined;
+          return {
+            ...story,
+            title,
```
rot1024, line 61:
```suggestion
            title: title || story.title,
```

rot1024 (approved):

PR #47, opened 2021-07-16
fix: infobox colors

issmail-basel:
# Overview
Infobox's title color cannot be changed

## What I've done
as agreed with @lavalse, change the infobox title color when color is set.
text block title color same as content.


lavalse:
Seems OK!  👍

rot1024 (dismissed):

PR #58, opened 2021-08-04
feat: draggable layer

HideBa:
# Overview
* Allowed users to change layer's location property by drag and drop

## What I've done
* Dev cesium-dnd package
* added it on Re:Earth
* When the visualizer is initialized, instantiate cesium-dnd
* When layers position have been changed, call mutation

## What I haven't done
* On the design, when a user is dragging the layer the mouse's lat lng should be displayed, but I haven't

## How I tested
* I've tested that I can change layers position by drag and drop
	* sphere
	* photo-overlay
	* marker
* I've checked that property value is also updated
* I've tested that if I change the linked layer, the location value is overwritten.

## Screenshot
6f7bc0054848242de3427181f89cff92

* If I don't break the responsibility of visualizer, engine, etc.
* Any tips to improve the performance of dnd


lavalse:
I guess all the layers also can be draggable in preview page and publish project page. So it should not work like that.

lavalse:
When test, I found this new function will make the cesium view become quite slow. Especially while rotating view of earth editor. I guess because cesium is testing user's behavior whether it is a drag layer or not. If you have time , I can show you what I found. After all I think this function still need to think about, and should not be merged ...

HideBa:
@lavalse 
>I guess all the layers also can be draggable in preview page and publish project page. So it should not work like that.

Thanks for letting me know that. 
I've just fixed it. Plz, check the preview page.

HideBa:
I've updated UI.
Plz review.

3e1cd0a7f50a83c19dbeff1dd8d4ea9e

rot1024:
Can we disable the screen turning light blue when dragging? It's a feature that has been around since the old Re:Earth, but I find it obtrusive anymore.

HideBa:
@rot1024 This is the UI I've talked with @lavalse since the app should return any feedback that the user is dragging. How do you think @lavalse ?

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -27,13 +30,21 @@ export default ({
...
   ref: React.ForwardedRef<EngineRef>;
   property?: SceneProperty;
   camera?: Camera;
   selectedPrimitiveId?: string;
   onPrimitiveSelect?: (id?: string, options?: SelectPrimitiveOptions) => void;
   onCameraChange?: (camera: Camera) => void;
+  isLayerDraggable?: boolean;
+  onDragLayer?: (layerId: string, position: LatLngHeight | undefined) => void;
+  onDraggingLayer?: (layerId: string, position: LatLngHeight | undefined) => void;
+  onDropLayer?: (layerId: string, position: LatLngHeight | undefined) => void;
```
KaWaite, line 47:
```suggestion
  onDragLayer?: (layerId: string, position?: LatLngHeight) => void;
  onDraggingLayer?: (layerId: string, position?: LatLngHeight) => void;
  onDropLayer?: (layerId: string, position¿: LatLngHeight) => void;
```

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -167,10 +178,60 @@ export default ({
...
+    if (!cartographic) return;
+    return {
+      lat: CesiumMath.toDegrees(cartographic.latitude),
+      lng: CesiumMath.toDegrees(cartographic.longitude),
+      height: cartographic.height,
+    };
+  };
+
+  //Enable Drag and Drop Layers
+  const DELAY = 100;
```
KaWaite, line 223: Sorry super small, but also wanted to hear your opinion on this.
Is this really necessary? I know if where it is being used is not clear doing this will make it clear what the value is. But below where this is used the key is `dragDelay`, which is clear even if it is `dragDelay: 100`.
Don't need to change but would like to hear if you think this is now overkill.

src/components/molecules/Visualizer/Engine/index.tsx
```diff
@@ -69,6 +69,11 @@ export type EngineProps = {
   primitiveSelectionReason?: string;
   onPrimitiveSelect?: (id?: string, options?: SelectPrimitiveOptions) => void;
   onCameraChange?: (camera: Camera) => void;
+  isLayerDraggable?: boolean;
+  onDragLayer?: (layerId: string, position: LatLngHeight | undefined) => void;
+  onDraggingLayer?: (layerId: string, position: LatLngHeight | undefined) => void;
+  onDropLayer?: (layerId: string, position: LatLngHeight | undefined) => void;
```
KaWaite, line 75:
```suggestion
  onDragLayer?: (layerId: string, position?: LatLngHeight) => void;
  onDraggingLayer?: (layerId: string, position?: LatLngHeight) => void;
  onDropLayer?: (layerId: string, position?: LatLngHeight) => void;
```

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
```diff
@@ -115,12 +115,10 @@ const Wrapper = styled.div`
   height: 80px;
   width: 500px;
   box-shadow: 0 4px 4px rgba(0, 0, 0, 0.25);
```
KaWaite, line 117: Any reason why you removed space on this file? Doesn't it make things harder to read? Pretty sure most components keep space between media queries, hovers, etc for clarity.

src/components/organisms/EarthEditor/CanvasArea/convert.ts
```diff
@@ -231,3 +231,55 @@ export const convertToBlocks = (data?: GetBlocksQuery): BlockType[] | undefined
     .filter((a): a is BlockType[] => !!a)
     .reduce((a, b) => [...a, ...b], []);
 };
+
+// TODO: refactor these part since it isn't based on DRY principle
+export const convertLayersWithRawProperty = (
+  data: GetLayersQuery | undefined,
```
KaWaite, line 237:
```suggestion
  data?: GetLayersQuery,
```

KaWaite: Similar places in this file also exist so update those as well to keep consistent.

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
```diff
@@ -201,19 +195,16 @@ const Menu = styled.div<{ menuOpen?: boolean }>`
   border-radius: ${metricsSizes["s"]}px;
   display: ${({ menuOpen }) => (!menuOpen ? "none" : "")};
```
issmail-basel, line 196: I think `unset` value is more appropriate than an empty string, like this:
`display: ${({ menuOpen }) => (!menuOpen ? "none" : "unset")};`

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
HideBa: I don't have a strong reason why I declared this as a variable. 
This is a kind of config value and I thought in the future we might move this variable to another config file if needed. To do that, I chose this.

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
HideBa: I didn't touch this file but the code formatter did this...
To follow the formatter's config, I thought I should keep this. 🤔

src/components/organisms/EarthEditor/CanvasArea/hooks.ts
```diff
@@ -140,6 +149,39 @@ export default (isBuilt?: boolean) => {
     document.title = title;
   }, [isBuilt, title]);
 
+  const [updateLayerLatLng] = useUpdatePropertyValueLatLngMutation();
+
+  const handleDragLayer = useCallback((_layerId: string, _position: LatLngHeight | undefined) => {
```
KaWaite, line 154: Let's keep style as consistent as we can!
```suggestion
  const handleDragLayer = useCallback((_layerId: string, _position?: LatLngHeight) => {
```

KaWaite, line 154: _layerId and _position aren't being used in this function. All this does is set dragging to true.

src/components/organisms/EarthEditor/CanvasArea/hooks.ts
```diff
@@ -140,6 +149,39 @@ export default (isBuilt?: boolean) => {
...
+    setIsDragging(true);
+  }, []);
+
+  const handleDropLayer = useCallback(
+    async (layerId: string, position: LatLngHeight | undefined) => {
+      setIsDragging(false);
+      const layerProperty = layers?.layers.find(l => l.id === layerId)?.rawProperty;
+      const fieldId = "location";
+      const schemaItemId = "default";
```
KaWaite, line 163: I feel like this callback is noisier than it needs to be. Can't we just set these in the updateLayerLatLng call? Rather than here writing `const fieldId = "location"`, below we can just pass it and schemaItemId in like `fieldId: "location"`. 
Again, my understanding is to do what you did only for clarity, which the param names updateLayerLatLng is expecting are clear enough (the exact same actually).

src/components/organisms/EarthEditor/CanvasArea/hooks.ts
```diff
@@ -140,6 +149,39 @@ export default (isBuilt?: boolean) => {
...
+        layerProperty && "items" in layerProperty
+          ? layerProperty?.items.find(
+              i => i.__typename === "PropertyGroup" && i.fields.find(f => f.fieldId === "location"),
+            )
+          : undefined;
+      if (!layerProperty?.id || !position) return;
+      await updateLayerLatLng({
+        variables: {
+          propertyId: layerProperty?.id,
+          itemId: propertyItem?.id,
```
KaWaite, line 174:
```suggestion
          propertyId: layerProperty.id,
          itemId: propertyItem.id,
```
Since you have the check on line 170 I don't think you need the question marks here. If you get a warning here, you can just update line 70 like this:
```
if (!layerProperty || !layerProperty.id || !position) return;

File: src/gql/graphql-client-api.tsx [hunk 10]
```diff
@@ -2413,6 +2413,37 @@ export type AddInfoboxFieldMutation = (
   )> }
 );
 
+export type UpdatePropertyValueLatLngMutationVariables = Exact<{
+  propertyId: Scalars['ID'];
+  schemaItemId?: Maybe<Scalars['PropertySchemaFieldID']>;
+  itemId?: Maybe<Scalars['ID']>;
```
KaWaite, line 2419: Are these actually not necessary?

src/components/molecules/Visualizer/Engine/Cesium/index.tsx
```diff
@@ -74,6 +83,7 @@ const Cesium: React.ForwardRefRenderFunction<EngineRef, EngineProps> = (
           width: small ? "300px" : "auto",
           height: small ? "300px" : "100%",
           display: ready ? undefined : "none",
+          cursor: isLayerDragging ? "grab" : "default",
```
KaWaite, line 86: Before adding this here, did you try to add the cursor effects on the layer itself? I feel like this is a little hacky and it would be amazing if the cursor could show feedback on hover of each layer (then the user will know somehting is possible in the correct place). 
Currently if I hover over a layer it stays the default, which makes it hard to click correctly on the layer.

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -167,10 +178,60 @@ export default ({
...
+    if (!cartographic) return;
+    return {
+      lat: CesiumMath.toDegrees(cartographic.latitude),
+      lng: CesiumMath.toDegrees(cartographic.longitude),
+      height: cartographic.height,
+    };
+  };
+
+  //Enable Drag and Drop Layers
+  const DELAY = 1000;
```
KaWaite, line 223: Can you play with this a little more or chat with the team? For me, 1 second feels a little long.

lavalse:
For me, I still feel the performance is not good enough after changing the maker's location, and the fps just come to less than 20 every time I rotate the earth. If we should test on a better back-end environment, just merge to main, see what will happen.
Thanks

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
lavalse: yesterday we set to less than 0.01s I guess, it will course some error when we select the marker, I think for now 1 s is ok. This needs more tests I think.

src/components/organisms/EarthEditor/CanvasArea/hooks.ts
HideBa: I think the interface of the component which are passed this function as props shouldn't change because of this implementation. The cesium-dnd expect to recieve this type of function and to keep the interface stable, I chose this way.
I assume `_` is conventionally used for ignored params.
@rot1024 How do you think about this?

src/gql/graphql-client-api.tsx
HideBa: This is the file generated automatically.

src/components/molecules/Visualizer/Engine/Cesium/index.tsx
HideBa: If we want to add the cursor effects on the layer itself, I assume we need to force `primitive` to have `onHover` props. We can't guarantee that all primitives will have that since they might be developed by external developers.

PR #64, opened 2021-08-11
fix: icon background

rot1024:
## What I've done

image

- Bounding box of Icon component has been fixed
- Now `<Icon src="<svg>...</svg>" />` works again
- Support style prop of Icon component


lavalse (approved):
Thanks very much

PR #67, opened 2021-08-23
feat: enable to set theme for the scene

issmail-basel:
# Overview
Add Scene theme option, users can define the basic theme which is used for all components on the scene.

## What I've done
- Add pre made theme colors
- Add option to change used theme/colors in scene tab
- Calculate colors for custom theme

## How I tested
- Change options in scene tab -> `Publish Theme`
- Choose `pre made theme` or `custom theme` with color preferences
- Check changes in the `Editor`, `Preview page` and `published project`

- Ways to improve code quality
- Any missing styles


issmail-basel:
@lavalse , what about menu colors, should they stay dark?
if not please update figma to take care of missing parts

lavalse:
@issmail-basel For the menu widgets, because we will refine the logic in the future. So for now, 
image
this is the solution. They should also use the theme by default I think. Is that ok?

src/components/molecules/Visualizer/Engine/index.tsx
```diff
@@ -53,6 +53,13 @@ export type SceneProperty = {
     enableGA?: boolean;
     trackingId?: string;
   };
+  reTheme?: {
+    reThemeMode?: "preMadeTheme" | "customTheme";
+    preMadeThemeId?: "reEarthLight" | "reEarthDark" | "forest";
+    textColor?: string;
+    selectColor?: string;
+    backgroundColor?: string;
+  };
```
rot1024, line 62:
```suggestion
  theme?: {
    themeType?: "light" | "dark" | "forest" | "custom";
    themeTextColor?: string;
    themeSelectColor?: string;
    themeBackgroundColor?: string;
  };
```

Refer to https://github.com/reearth/reearth-backend/pull/35#discussion_r694462910

rot1024 (changes requested):
This is a somewhat roundabout implementation:

- Current: SceneProperty in Visualizer --> Global theme context --> Visualizer --> Each component (What was inside goes outside and then back inside again) 
- Should be: Visualizer (SceneProperty) --> Each component

Since public themes are used only inside Visualizer, they do not depend on the global context and should be used by defining a premade theme inside Visualizer

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
```diff
@@ -98,13 +101,23 @@ const InfoBox: React.FC<Props> = ({
           justify={open ? "flex-start" : "space-evenly"}
           direction="column"
           onClick={handleOpen}>
-          {isSmallWindow && !noContent && <StyledIcon icon="arrowUp" size={24} open={open} />}
+          {isSmallWindow && !noContent && (
+            <StyledIcon publishedTheme={publishedTheme} icon="arrowUp" size={24} open={open} />
+          )}
           <Text size="m" weight="bold" customColor>
-            <TitleText>{title || " "}</TitleText>
+            <TitleText publishedTheme={publishedTheme}>{title || " "}</TitleText>
```
KaWaite, line 108: Is this needed? Since the wrapper is getting the publishedTheme styles and the Text component that is wrapping this span is set to `customColor`, I feel like this shouldn't be needed.
If it is needed because the span isn't allowing the styles to be used let me know.

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
```diff
@@ -156,20 +169,22 @@ const TitleFlex = styled(Flex)`
   width: 75%;
 `;
 
-const StyledIcon = styled(Icon)<{ open?: boolean }>`
+const StyledIcon = styled(Icon)<{ open?: boolean; publishedTheme: PublishTheme }>`
```
KaWaite, line 172: This publishedTheme param is possibly undefined. Please fix it with `publishedTheme?: PublishTheme` and update how it is used for color.

src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
```diff
@@ -38,10 +38,12 @@ export type Props = {
   button: Button;
   menuItems?: MenuItem[];
   pos: Position;
+  sceneProperty?: any;
```
KaWaite, line 41:
```suggestion
  sceneProperty?: SceneProperty;
```

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
issmail-basel: Totally right, it's a mistake I made after resolving the conflicts.
Will fix.

src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
KaWaite: Also needs the SceneProperty imported

src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
```diff
@@ -153,28 +165,29 @@ const StyledIcon = styled(Icon)<{ margin: boolean }>`
   margin-right: ${({ margin }) => (margin ? "5px" : null)};
 `;
 
-const MenuWrapper = styled.div<{ visible?: boolean }>`
+const MenuWrapper = styled.div<{ visible?: boolean; publishedTheme: PublishTheme }>`
```
KaWaite, line 168: Again, publishedTheme is potentially undefined.

src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
issmail-basel: Sure

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
issmail-basel: `const { publishedTheme } = usePublishedTheme(sceneProperty?.theme);`
It is always defined and has a value, and returned from the hook.

src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
KaWaite: Same with the next three constants

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
issmail-basel: Even if param of the hook is undefined, it returns default values for `publishedTheme`.
So it is guaranteed to have a value.

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
```diff
@@ -130,19 +149,19 @@ const Wrapper = styled.div`
   }
 `;
 
-const ArrowButton = styled.button`
-  background-color: ${props => props.theme.main.paleBg};
+const ArrowButton = styled.button<{ publishedTheme: PublishTheme }>`
+  color: ${({ publishedTheme }) => publishedTheme.mask};
```
KaWaite, line 153: Is passing publishedTheme into all the styled components necessary? For example, this ArrowButton had `color: inherit` but you replaced it with the `publishedTheme.mainIcon`.
If you can check if everything inherits from the Wrapper's color attribute that would clean up this code a lot.

src/theme/publishTheme/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+  themeType?: "light" | "dark" | "forest" | "custom";
+  themeTextColor?: string;
+  themeSelectColor?: string;
+  themeBackgroundColor?: string;
+};
+
+export default (seneThemeOptions?: ReTheme) => {
+  const [publishedTheme, setPublishedTheme] = useState<PublishTheme>(dark);
+  function addAlpha(color?: string, opacity?: number): string {
+    // coerce values so ti is between 0 and 1.
```
KaWaite, line 17:
```suggestion
    // coerce values so they are between 0 and 1.
```

src/theme/publishTheme/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+import { PublishTheme } from "./publishTheme";
+
+export type ReTheme = {
+  themeType?: "light" | "dark" | "forest" | "custom";
+  themeTextColor?: string;
+  themeSelectColor?: string;
+  themeBackgroundColor?: string;
+};
+
+export default (seneThemeOptions?: ReTheme) => {
```
KaWaite, line 14: seneThemeOptions => sceneThemeOptions

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
issmail-basel: Unfortunately, every component is using different set of styles...
I have checked again and there is no way to inherit because child may using `mainIcon` and parent is using `mainText` etc...

src/theme/publishTheme/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+  const [publishedTheme, setPublishedTheme] = useState<PublishTheme>(dark);
+  function addAlpha(color?: string, opacity?: number): string {
+    // coerce values so ti is between 0 and 1.
+    const _opacity = Math.round(Math.min(Math.max(opacity || 1, 0), 1) * 255);
+    return color + _opacity.toString(16).toUpperCase();
+  }
+
+  function isDark(color: string) {
```
KaWaite, line 22: Check `atoms/Markdown/index.tsx` because it is using the tinycolor package already to check if a color is dark or not. 
Also check that package if it can be used to do any of the other things you are doing here manually!

src/theme/publishTheme/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+          background: seneThemeOptions.themeBackgroundColor,
+          mainText: seneThemeOptions.themeTextColor,
+          select: seneThemeOptions.themeSelectColor,
+          strongIcon: "#FFFFFF",
+          strongText: "#FFFFFF",
+          weakText: addAlpha(seneThemeOptions.themeTextColor, 0.5),
+          mainIcon: addAlpha(seneThemeOptions.themeTextColor, 0.5),
+          weakIcon: addAlpha(seneThemeOptions.themeTextColor, 0.25),
+        });
+      }
```
KaWaite, line 67: You could tidy this up a lot with a switch statement defaulting to dark.
Something like:
```
switch(sceneThemeOptions?.themeType){
  case "light":
    setPublishedTheme(light)
    break;
  case "forest"...
...
..
  default:
   setPublishedTheme(dark)
```

src/theme/styled.ts
```diff
@@ -7,3 +7,4 @@ import { Theme } from "./theme";
 
 export const styled = rawStyled;
 export const useTheme = () => useContext(ThemeContext) as Theme;
+export const SceneThemeUpdateContext = createContext<any>(undefined);
```
KaWaite, line 10: Is this being used? I can't find where you are using this so if it isn't being used please remove it

KaWaite (commented):
Left some comments! Let me know when you get a chance to check them/update the PR.

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
KaWaite: If publishedTheme is always defined then update this component's Props (and others that apply). You have it as a conditional prop above.

src/theme/publishTheme/hooks.ts
```diff
@@ -0,0 +1,62 @@
+import { useEffect, useState } from "react";
+import tinycolor from "tinycolor2";
+import { dark } from "./darkTheme";
+import { forest } from "./forestTheme";
+import { light } from "./lightTheme";
+import { PublishTheme } from "./publishTheme";
+
+export type SeneThemeOptions = {
```
KaWaite, line 8: This still isn't updated.
Should be `SceneThemeOptions`

src/theme/publishTheme/hooks.ts
KaWaite: Still doesn't seem to be updated.

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
```diff
@@ -174,22 +194,23 @@ const Title = styled(Text)`
   }
 `;
 
-const StyledIcon = styled(Icon)`
-  color: ${props => props.theme.main.text};
+const StyledIcon = styled(Icon)<{ publishedTheme: PublishTheme }>`
```
KaWaite, line 197: Sorry, one more thing. In components like this StyledIcon where all you are using is mainIcon from publishedTheme, can you pass JUST what you need, rather than the whole publishedTheme object.
In this case you could pass a `color=publishedTheme.mainIcon` prop instead

src/components/molecules/Visualizer/Infobox/Frame/index.tsx
issmail-basel: Please explain more, I can't figure out what you mean?
It is not conditional as it always passed and has a value

KaWaite: Sorry my bad. I think I was mixing up the sceneProperty prop and the publishedTheme...!

src/components/molecules/Visualizer/Widget/Menu/MenuButton.tsx
KaWaite: This still isn't resolved.

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
KaWaite: This wasn't addressed. Please let me know if you don't catch what I mean and I can explain more!

src/theme/publishTheme/hooks.ts
KaWaite: Not resolved

src/theme/publishTheme/hooks.ts
KaWaite: Still not resolved

src/theme/styled.ts
KaWaite: Please let me know about this.

src/components/molecules/Visualizer/Widget/Storytelling/index.tsx
```diff
@@ -160,7 +179,8 @@ const Current = styled(Flex)`
   }
 `;
 
-const Title = styled(Text)`
+const Title = styled(Text)<{ color: string }>`
+  color: ${({ color }) => color}!important;
```
KaWaite, line 183: The `!important` isn't necessary. Above where you use Title just pass `customColor` and you should be good to go!
ie
```
<Title color={publishedTheme.mainText} size="m" weight="bold" customColor>
....
```

KaWaite (approved):
Just one change requested and then LGTM! I'll approve but please make the change!

rot1024 (changes requested):
`usePublicTheme` is redundant, and I don't see the need to use Hooks. It's enough to simply have the following code:

`@reearth/theme/public.ts`:

```ts
export type Theme = { backgroundColor: string, /*...*/ };
export type ThemeType = "dark" | "light" | "forest";
export type ThemeConfig = Partial<Theme> & { type: ThemeType | "custom" };

const defaultTheme: ThemeType = "dark";
const premade: {[key in ThemeType]: Theme } = {
  dark: { ... },
  light: { ... },
  forest: { ... }
};

export function publicTheme(theme?: ThemeConfig): Theme {
  return theme?.type === "custom" ? { ...premade[defaultTheme], ...theme } : premade[theme?.type || defaultTheme];
}
```

Usage:

```tsx
import { publicTheme } from "@reearth/theme";

function Storytelling({ sceneProperty }) {
  const theme = publicTheme(sceneProperty?.theme);

  // use theme!
}
```

This code doesn't use hooks so the code is very clean.

issmail-basel:
@rot1024 , thank you for your comment.
but hooks are needed to detect theme changes.
Your code is working when the component is rendered for the first time, but if we change the pre-made theme to (light for example) it won't detect the change.
That's why I've used hooks (`useEffect`).
What do you recommend to solve this issue? Should I keep using hooks?

rot1024:
It's strange. In my code, theme should switch when sceneProperty will be updated. Let me try my code in this PR.

rot1024:
OK, and I tried and I understand hooks is needed because custom theme uses object and it's needed to prevent useless re-rendering. But there are rooms to be refactored so I did.

Please check:

https://github.com/reearth/reearth-web/compare/scene-theme-front...scene-theme-front-refactor

Points:

- `useState` and `useEffect` can be replaced with `useMemo`
- Non-unified name: PublishTheme and PublishedTheme, I unified them with PublishTheme
- Simpler file name and fewer files at `src/theme`

rot1024 (approved):

PR #72, opened 2021-09-02
refactor: replace deprecated gql fields, pass widgetId to widget mutations

rot1024:
# Overview

## What I've done

Follow https://github.com/reearth/reearth-backend/pull/43 's change:

- Replace deprecated fields in queries
- Change input of widget mutation

No UI changes

## How I tested

Manually tested

### DIFF
--- a/src/components/organisms/EarthEditor/OutlinePane/hooks.tsx
+++ b/src/components/organisms/EarthEditor/OutlinePane/hooks.tsx
@@ -78,12 +78,15 @@ export default () => {
           .map((e): Widget => {
             const pluginId = plugin.id;
             const extensionId = e.extensionId;
-            const enabled = scene?.widgets.find(
+            // note: multiple widget is not supported now
+            const widget = scene?.widgets.find(
               w => w.pluginId === plugin.id && w.extensionId === e.extensionId,
-            )?.enabled;
+            );
             return {
-              id: `${plugin.id}/${e.extensionId}`,
-              enabled,
+              id: widget?.id,
+              pluginId,
+              extensionId: e.extensionId,
+              enabled: !!widget?.enabled,
               title: e.translatedName,
               description: e.translatedDescription,
               icon: e.icon || (pluginId === "reearth" ? extensionId : undefined),
@@ -123,8 +126,13 @@ export default () => {
   }, [select, selectBlock]);
 
   const selectWidget = useCallback(
-    (id: string) => {
-      select({ type: "widget", widgetId: id });
+    (widgetId: string | undefined, pluginId: string, extensionId: string) => {
+      select({
+        type: "widget",
+        widgetId,
+        pluginId,
+        extensionId,
+      });
       selectBlock(undefined);
     },
     [select, selectBlock],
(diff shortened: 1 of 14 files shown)


HideBa (approved):
LGTM

src/components/organisms/EarthEditor/PropertyPane/hooks-queries.ts
```diff
@@ -144,14 +140,14 @@ export default ({
...
-          w.pluginId === selectedWidgetId.pluginId &&
-          w.extensionId === selectedWidgetId.extensionId,
-      );
+      if (selected?.type !== "widget") return;
+      const w = selected.widgetId
+        ? scene?.widgets.find(w => selected.widgetId === w.id)
+        : undefined;
       return {
-        id: selectedWidgetId.pluginId + "/" + selectedWidgetId.extensionId,
+        id: selected.widgetId || `${selected.pluginId}/${selected.widgetId}`,
```
KaWaite, line 148:
```suggestion
        id: selected.widgetId || `${selected.pluginId}/${selected.extensionId}`,
```
This is a mistake, right?

src/components/organisms/EarthEditor/PropertyPane/hooks.ts
```diff
@@ -39,21 +39,14 @@ export type FieldPointer = {
 };
 
 export default (mode: Mode) => {
-  const [selected] = useSelected();
+  const [selected, select] = useSelected();
```
KaWaite, line 42:
```suggestion
  const [selected] = useSelected();
```
I don't think the select is being used.

src/components/organisms/EarthEditor/OutlinePane/hooks.tsx
```diff
@@ -245,7 +253,8 @@ export default () => {
     sceneDescription,
     selectedType: selected?.type,
     selectedLayerId: selected?.type === "layer" ? selected.layerId : undefined,
-    selectedWidgetId: selected?.type === "widget" ? selected.widgetId : undefined,
+    selectedWidgetId:
+      selected?.type === "widget" ? `${selected.pluginId}/${selected.extensionId}` : undefined,
```
KaWaite, line 257: I know we aren't supporting multi-widgets yet, but is there a reason why you are using `pluginId/extensionId` instead of just `widgetId` OR `widgetId/pluginId/extensionId`?
I could see we use this `selectedWidgetId` inside the **TreeView** so if we are using this value on updating `selectedWidget` shouldn't we be using all three?

KaWaite (commented):
Just a couple points I wanted clarified in case there was a mistake!

src/components/organisms/EarthEditor/PropertyPane/hooks-queries.ts
rot1024: Thanks!

src/components/organisms/EarthEditor/OutlinePane/hooks.tsx
rot1024: Widgets (note: extensions, not scene widgets) are always displayed on the left panel even when there are no widgets on the scene so no widget ID exists. For now multiple widget is supported only by back-end, not front-end. To identify widgets in that case, we have to treat a set of pluginId and extensionId.

KaWaite: Gotcha!

KaWaite (approved):
Looks good!

PR #73, opened 2021-09-06
feat: Notification system update

KaWaite:
# Overview
This is an update to the notification system. We were creating notifications in a couple different ways and had duplicate code, so this cleans things up and is trying to make things simpler.

## What I've done
- Moved notification code into its own hooks file and created a notification wrapper component that sits above the router so it doesn't refresh on page changes.
- Added more notifications + updated translations

## What I haven't done
- GQL errors coming from the backend aren't often user-facing appropriate. Backend error handling still needs to be sorted out eventually

## How I tested
- Did my best to test all notifications. *Some errors I couldn't test for one reason or another.*

## Screenshot
Screen Shot 2021-09-06 at 16 26 27
Screen Shot 2021-09-06 at 16 26 40
Screen Shot 2021-09-06 at 16 27 00

- Translations (@rot1024 @HideBa)
- Where the notification code should go (I wasn't sure the best place, but since it wraps the whole app similar to the various providers, I thought maybe it should be where I put it in the src directory)

## test cases
Dashboard
- Create project
- Create workspace
Settings
- change password
- create project
- archive project
- delete project
- create workspace
- delete workspace
Earth Editor
- Publish project
- Unpublish project
- Add dataset


src/notifications/README.md
```diff
@@ -1,8 +1,8 @@
-# NotificationBar
+# NotificationBanner
 [![Generic badge](https://img.shields.io/badge/GROUP-global-<COLOR>.svg)]()
 [![Generic badge](https://img.shields.io/badge/SIZE-atom-blue.svg)]()
 
-NotificationBar. This component is a global component.
+NotificationBanner. This component is a global component.
 
 ## Usage
```
rot1024, line 7: README can be deleted because it is not used

src/notifications/hooks.ts
```diff
@@ -0,0 +1,59 @@
...
+import { useNotification, Notification } from "@reearth/state";
+
+export type NotificationStyleType = "error" | "warning" | "info" | "success";
+export type NotificationType = Notification;
+
+export default () => {
+  const intl = useIntl();
+  const [notification, setNotification] = useNotification();
+  const [visible, changeVisibility] = useState(false);
+  const notificationTimeout = 5000;
```
rot1024, line 12: This lien can be moved outside the function

src/notifications/hooks.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    switch (notification.type) {
+      case "error":
+        return intl.formatMessage({ defaultMessage: "Error" });
+      case "warning":
+        return intl.formatMessage({ defaultMessage: "Warning" });
+      default:
+        return intl.formatMessage({ defaultMessage: "Notice" });
+    }
+  }, [intl, notification]);
```
rot1024, line 26: Messages are static value so you don't need to use useMemo:


```suggestion
  const errorMessage = intl.formatMessage({ defaultMessage: "Error" });
  const warningMessage = intl.formatMessage({ defaultMessage: "Warning" });
  const noticeMessage = intl.formatMessage({ defaultMessage: "Notice" });

  const notificationHeading = {
    error: errorMessage
    warning: warningMessage
  }[notification?.type] || noticeMessage;
```

src/notifications/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              setNotification(undefined);
+            }}
+          />
+        </HeadingArea>
+        <Text size="s" color={theme.notification.text}>
+          {notification?.text}
+        </Text>
+      </StyledNotificationBanner>
+      {children}
+    </div>
```
rot1024, line 42: This component doesn't need to have children, otherwise the component may become too complex:

```suggestion
  <StyledNotificationBanner visible={visible} type={notification?.type} direction="column">
    <HeadingArea justify="space-between">
      <Text
        size="m"
        color={theme.notification.text}
        weight="bold"
        otherProperties={{ padding: "0 0 8px 0" }}>
        {notification?.heading}
      </Text>
      <CloseBtn
        icon="cancel"
        size={20}
        onClick={() => {
          changeVisibility(false);
          setNotification(undefined);
        }}
      />
    </HeadingArea>
    <Text size="s" color={theme.notification.text}>
      {notification?.text}
    </Text>
  </StyledNotificationBanner>
```

src/app.tsx
```diff
@@ -42,24 +43,26 @@ const App: React.FC = () => {
...
+                    <SettingsProjectList path="/settings/workspace/:teamId/projects" />
+                    <AssetSettings path="/settings/workspace/:teamId/asset" />
+                    <ProjectSettings path="/settings/project/:projectId" />
+                    <PublicSettings path="/settings/project/:projectId/public" />
+                    <DatasetSettings path="/settings/project/:projectId/dataset" />
+                    <PluginSettings path="/settings/project/:projectId/plugins" />
+                    {process.env.NODE_ENV !== "production" && <GraphQLPlayground path="/graphql" />}
+                    <NotFound default />
+                  </StyledRouter>
+                </NotificationBanner>
```
rot1024, line 65: Check the comment at Notifications component, and then this code is enough:

```suggestion
                <NotificationBanner />
                <StyledRouter>
                  <TopPage path="/" />
                  <Dashboard path="/dashboard/:teamId" />
                  <EarthEditor path="/edit/:sceneId" />
                  <Preview path="/edit/:sceneId/preview" />
                  <Redirect from="/settings" to="/settings/account" />
                  <AccountSettings path="/settings/account" />
                  <WorkspaceList path="/settings/workspaces" />
                  <WorkspaceSettings path="/settings/workspace/:teamId" />
                  <SettingsProjectList path="/settings/workspace/:teamId/projects" />
                  <AssetSettings path="/settings/workspace/:teamId/asset" />
                  <ProjectSettings path="/settings/project/:projectId" />
                  <PublicSettings path="/settings/project/:projectId/public" />
                  <DatasetSettings path="/settings/project/:projectId/dataset" />
                  <PluginSettings path="/settings/project/:projectId/plugins" />
                  {process.env.NODE_ENV !== "production" && <GraphQLPlayground path="/graphql" />}
                  <NotFound default />
                </StyledRouter>
```

src/app.tsx
```diff
@@ -9,6 +9,7 @@ import { Provider as Auth0Provider } from "./auth";
 
 import NotFound from "@reearth/components/pages/NotFound";
 import Loading from "@reearth/components/atoms/Loading";
+import NotificationBanner from "@reearth/notifications";
```
rot1024, line 12:
```suggestion
import NotificationBanner from "@reearth/components/organisms/Nnotification";
```

because notifications module doesn't provide any context (provider), so this component should follow [the Component Classification Guideline](https://github.com/reearth/reearth-web/blob/main/src/components/README.md)

This component is not a page, but uses `@reearth/state`, so I think this component should be classified as a organism.

rot1024 (changes requested):
Let's try to write a component unit test!

src/notifications/hooks.ts
KaWaite: The above code doesn't work for a couple different reasons: notification.type is typed and won't allow JUST error and warning (it expects options for *info* and *success* as well). Also, since *notification* is possibly undefined, it doesn't like that.
I tried a couple other approaches but I can't find anything that works/works well like the useMemo.
Any other advice to keep this clean instead of me doing something like
```
const notificationHeading = notification?.type === "warning" ? intl.... : notification.type === "error" ? ..... etc etc
```

src/components/pages/TopPage/hooks.ts
```diff
@@ -2,7 +2,7 @@ import { useEffect } from "react";
 import { useNavigate } from "@reach/router";
 
 import { useAuth, useCleanUrl } from "@reearth/auth";
-import useNotification from "@reearth/notifications/hooks";
+import useNotification from "@reearth/components/organisms/Notification/hooks";
```
rot1024, line 5:
```suggestion
import { useNotification } from "@reearth/state";
```

`@reearth/components/organisms/Notification/hooks` cannot be used directly by external components.

Directly using hooks of the component by external components complicates the dependencies between components and is therefore prohibited. 

Don't forget that we have gone to the trouble of introducing jotai to prevent this from happening. Otherwise, jotai itself is unnecessary.

src/notifications/hooks.ts
rot1024: > notification.type is typed and won't allow JUST error and warning (it expects options for info and success as well). Also, since notification is possibly undefined, it doesn't like that.

Currently there are only 3 messages for notification bar, so I think it is not a problem. Types except error and warning will become undefined, but then it will become "notice" message thanks to `|| noticeMessage`.

Also, an implementation using ternary operators is also acceptable.

rot1024 (changes requested):
Would you like to write unit tests for Notification component?

src/notifications/hooks.ts
KaWaite: Sorry, I should have made clear I was getting type errors for the above code, even if I modified it a bit. So Typescript wasn't allowing me to use the above code. 
I'll go with a ternary operator. Thank you

src/components/pages/TopPage/hooks.ts
KaWaite: This is actually my poor naming. `useNotification` here is the name I gave for the notification-related hooks. As you can see below on line 16 I pull out notify, which is the function being used to update the notification (with checks and adding the notificationHeading).
Is there a more appropriate name I should use? I didn't think calling it `useHooks` was appropriate since that is the name we use for local hooks use.
Let me know if there is a better name since useNotification can get confused with Jotai.

rot1024: I think it's not about the name, it's about the fact that notifications from each page and organisms should always be updated via jotai, and useHooks should not be imported from any component other than organims/Notification. That hooks used internally.

src/components/molecules/Common/Notification/index.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+
+import Flex from "@reearth/components/atoms/Flex";
+import Icon from "@reearth/components/atoms/Icon";
+import Text from "@reearth/components/atoms/Text";
+import { styled, metrics, useTheme } from "@reearth/theme";
+
+export type NotificationStyleType = "error" | "warning" | "info" | "success";
+export type NotificationType = {
+  type: "error" | "warning" | "info" | "success";
```
rot1024, line 10:
```suggestion
  type: NotificationStyleType;
```

src/components/molecules/Common/Notification/index.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+
+import Flex from "@reearth/components/atoms/Flex";
+import Icon from "@reearth/components/atoms/Icon";
+import Text from "@reearth/components/atoms/Text";
+import { styled, metrics, useTheme } from "@reearth/theme";
+
+export type NotificationStyleType = "error" | "warning" | "info" | "success";
+export type NotificationType = {
```
rot1024, line 9:
```suggestion
export type Notification = {
```

src/components/molecules/Common/Notification/index.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+
+import Flex from "@reearth/components/atoms/Flex";
+import Icon from "@reearth/components/atoms/Icon";
+import Text from "@reearth/components/atoms/Text";
+import { styled, metrics, useTheme } from "@reearth/theme";
+
+export type NotificationStyleType = "error" | "warning" | "info" | "success";
```
rot1024, line 8:
```suggestion
export type NotificationType = "error" | "warning" | "info" | "success";
```

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/hooks.ts
```diff
@@ -16,7 +16,7 @@ export default (
     sheetName: string,
     schemeId: string | null,
   ) => Promise<void>,
-  onNotify?: (type: NotificationType, text: string) => void,
+  onNotify?: (type: NotificationStyleType, text: string) => void,
```
rot1024, line 19:
```suggestion
  onDatasetAdd?: () => void;
```

Notification should be implemented at organisms. Molecules should not handle anything about notification.

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/hooks.ts
```diff
@@ -31,16 +31,13 @@ export default (
       await handleGoogleSheetDatasetAdd(sheet.accessToken, sheet.fileId, sheet.sheetName, null);
       onNotify?.(
         "success",
-        intl.formatMessage({ defaultMessage: "You have added a dataset successfully." }),
+        intl.formatMessage({ defaultMessage: "Successfully added a dataset!" }),
       );
```
rot1024, line 35:
```
      onDatasetAdd?.()
```

src/components/molecules/EarthEditor/DatasetPane/DatasetModal/hooks.ts
```diff
@@ -31,16 +31,13 @@ export default (
...
     }
     const data = dataType === "csv" ? csv : url;
     if (!data || !handleDatasetAdd) return;
     await handleDatasetAdd(data, null);
-    onNotify?.(
-      "success",
-      intl.formatMessage({ defaultMessage: "You have added a dataset successfully." }),
-    );
+    onNotify?.("success", intl.formatMessage({ defaultMessage: "Successfully added a dataset!" }));
```
rot1024, line 40:
```suggestion
    onDatasetAdd?.(};
```

src/components/molecules/EarthEditor/PublicationModal/hooks.ts
```diff
@@ -13,7 +13,7 @@ export type CopiedItemKey = {
 export default (
   defaultAlias?: string,
   onClose?: () => void,
-  onNotify?: (type: NotificationType, text: string) => void,
+  onNotify?: (type?: NotificationStyleType, text?: string) => void,
```
rot1024, line 16:
```suggestion
  onCopyToClipBoard?: () => void,
```

src/components/molecules/EarthEditor/PublicationModal/hooks.ts
```diff
@@ -48,8 +48,8 @@ export default (
           "info",
           `${
             key === "embedCode"
-              ? intl.formatMessage({ defaultMessage: "Embed code was successfully copied!" })
-              : intl.formatMessage({ defaultMessage: "URL was successfully copied!" })
+              ? intl.formatMessage({ defaultMessage: "Successfully copied the embed code!" })
+              : intl.formatMessage({ defaultMessage: "Successfully copied the URL!" })
           }`,
```
rot1024, line 53:
```ts
onCopyToClipBoard?.();
```

rot1024 (changes requested):
Molecules should not be responsible for notifications; please run notifications in organisms. Otherwise, the dependencies between components will become too complex.

src/components/molecules/Common/Notification/index.tsx
KaWaite: NotificationType comes from `@reearth/state` and we have it set up so I can't import from there. I thought about exporting it from `organisms/NotificationBanner` but then I'd have looping imports which isn't good.
Any other options or can I leave this the way it is?

rot1024: You can leave it. In thi case, it is not a problem to have two types. If one of the types is changed and the types are no longer aligned, an error will occur and you'll know immediately.

src/components/molecules/Common/Notification/index.test.tsx
```diff
@@ -2,9 +2,9 @@ import React from "react";
 
 import { render, screen } from "@reearth/test/utils";
 
-import NotificationBanner, { NotificationType } from "./index";
+import NotificationBanner, { Notification } from "./index";
```
rot1024, line 5:
```suggestion
import NotificationBanner, { Notification } from ".";
```

rot1024 (approved):

src/components/molecules/Common/Notification/index.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+
```
HideBa, line 2: Should this component be a molecule of our system?
I assume this component doesn't have any knowledge about Re:Earth and should be an atom.

HideBa (commented):
I quickly reviewed and left a comment.

src/components/molecules/Common/Notification/index.tsx
KaWaite: It is using multiple atom components which we don't explicitly prohibit (I don't think), but I feel like we implicitly prohibit it since atoms should be the "smallest", most basic piece. Whereas molecules are built up of atoms and other molecules.
So I will argue that it is exactly where it should be ;)

HideBa: Based on the typical atomic design principle, I agree that this is a molecule. But our component classification rules, atoms can depend on atoms and the difference between atoms and molecules is if they know the existence of Re:Earth.  I assume this component is more general and can be used in other apps as well so I thought it's an atom.
@rot1024 How do you think?

KaWaite: Turns out it is an Atom

HideBa (approved):
LGTM!

rot1024 (approved):

PR #77, opened 2021-09-09
fix: password updating

KaWaite:
# Overview
Users were unable to change their password

## What I've done
The frontend's handling of updating the password was wrong. We were sending to the backend the old password and the new password confirmation. We should send the new password and confirmation only. 
Anyways the user doesn't need to input their old password bc they are authenticated. 
Also found a warning in AccountSection so cleaned up that component.
Added type prop to TextBox so we could use it for passwords and hide the inputted characters.

## How I tested
- Changed my own password on local multiple times

### DIFF
--- a/src/components/molecules/Settings/Account/AccountSection/index.tsx
+++ b/src/components/molecules/Settings/Account/AccountSection/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useState, useMemo } from "react";
 import PasswordModal from "@reearth/components/molecules/Settings/Account/PasswordModal";
 import Section from "@reearth/components/molecules/Settings/Section";
 import EditableItem from "@reearth/components/molecules/Settings/Project/EditableItem";
@@ -8,6 +8,8 @@ import { styled } from "@reearth/theme";
 import { useIntl } from "react-intl";
 import Text from "@reearth/components/atoms/Text";
 import Flex from "@reearth/components/atoms/Flex";
+
+export type Theme = "DARK" | "LIGHT";
 
 export type Props = {
   email?: string;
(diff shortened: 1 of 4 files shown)


src/components/molecules/Settings/Account/AccountSection/index.tsx
```diff
@@ -34,22 +36,32 @@ const ProfileSection: React.FC<Props> = ({
...
-    const lang = items.find(item => item.key === intl.locale);
-    setCurrentLang(lang?.label);
-  }, [intl.locale]);
-
-  useEffect(() => {
-    const label = themeItems.find(themeItem => themeItem.key === appTheme)?.label;
-    setCurrentThemeLabel(label);
-  }, [appTheme]);
+
+  const themeItems: { key: Theme; label: string; icon: string }[] = useMemo(
```
HideBa, line 40: I don't have confident, do we need to memorize this as well.
Since it's just a variable and any calculation isn't needed. I assume this might take the memory more that without `memo`.

src/components/molecules/Settings/Account/PasswordModal/index.tsx
```diff
@@ -27,17 +27,24 @@ type Props = {
...
   const [disabled, setDisabled] = useState(true);
 
+  const handleClose = useCallback(() => {
+    setPassword("");
+    setPasswordConfirmation("");
+    onClose?.();
+  }, [onClose]);
+
   const save = useCallback(() => {
     if (password === passwordConfirmation) {
```
HideBa, line 43: If `password` doesn't match `passwordConfirmation` front-end should show the validation error inside of the form. 
Does this modal have any validation?

HideBa (changes requested):
Thank you for fixing the critical bug quickly. I left some comments. Kindly check. 👍

src/components/molecules/Settings/Account/PasswordModal/index.tsx
KaWaite: None yet. I was treating this as a hotfix since you said users already have passwords they want to change. It also doesn't have any regex to make sure it is of length, etc etc which I think eventually it should have. 
The reason I didn't do in this PR is because I am not comfortable with validation/regex and knew it would take me time to look into options/how to do it. 
Let me know if you want that included in this PR.

src/components/molecules/Settings/Account/AccountSection/index.tsx
KaWaite: My bad, I was thinking since it relies on internationalization wrapping it in a useMemo was good but now that I think about it we never do that with intl (that package updates itself correctly) :P I'll remove the useMemo.

src/components/molecules/Settings/Account/PasswordModal/index.tsx
HideBa: @KaWaite OK thank you. 👍 
You don't need to do that as a part of this PR.
Plz, create the issue `Add client-side validation on password modal` and write the description.
Maybe it'll be good for the first issue for contributers.

src/components/molecules/Settings/Account/AccountSection/index.tsx
KaWaite: Ah, my bad. I forgot I wrapped this in a useMemo because I was getting an eslint warning:
```
The 'themeItems' array makes the dependencies of useMemo Hook (at line 47) 
change on every render. To fix this, wrap the initialization of 'themeItems' in its 
own useMemo() Hook.eslint
```
For now I'll leave it in the useMemo!

HideBa: OK!

HideBa (approved):

rot1024 (approved):

PR #84, opened 2021-09-14
fix: language change not updating

KaWaite:
# Overview
Some parts of the UI wouldn't update after the language was changed. In particular the right panel when navigating from the Account Settings back to the Earth Editor.

## What I've done
Since the right panel's fields come from the backend and are stored in cache, after language is changed the apollo cache is reset so any possible translations coming from the backend will be fetched again with the new language.


rot1024 (approved):

HideBa (approved):
LGTM

PR #85, opened 2021-09-14
fix: showing members section for personal workspace

KaWaite:
# Overview
Had an issue where a personal workspace showed the members section when it shouldn't be rendered (personal workspace's only member is you, so no need).

## What I've done
- Cleanup code around updating team state
- fix where personal workspace page showed members section

## How I tested
- In the Dashboard, change to different workspace from header. Change back to personal workspace. Select "Manage workspaces". Click on workspace from left navigation. Should not show members section.

### DIFF
--- a/src/components/organisms/EarthEditor/Header/hooks.ts
+++ b/src/components/organisms/EarthEditor/Header/hooks.ts
@@ -85,11 +85,7 @@ export default () => {
     const team = teams?.find(t => t.id === teamId);
     if (!team) return;
 
-    setTeam({
-      id: team.id,
-      name: team.name,
-      personal: team.personal,
-    });
+    setTeam(team);
   }, [teams, currentTeam, teamId, setTeam]);
 
   useEffect(() => {
--- a/src/components/organisms/EarthEditor/Header/hooks.ts
+++ b/src/components/organisms/EarthEditor/Header/hooks.ts
@@ -88,6 +88,7 @@ export default () => {
     setTeam({
       id: team.id,
       name: team.name,
+      personal: team.personal,
     });
   }, [teams, currentTeam, teamId, setTeam]);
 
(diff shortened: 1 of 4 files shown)


HideBa (dismissed):
LGTM

src/components/organisms/EarthEditor/Header/hooks.ts
```diff
@@ -88,6 +88,7 @@ export default () => {
     setTeam({
```
issmail-basel, line 88: When using `SetTeam` why not passing the whole team object as you did in src/components/organisms/Dashboard/hooks.ts (for consistency)

src/components/organisms/Settings/SettingPage/hooks.ts
```diff
@@ -55,33 +55,12 @@ export default (params: Params) => {
   const team = teams?.find(team => team.id === teamId);
```
issmail-basel, line 55: Just a question.
Can this line be removed? and remove `team` from `useEffect` dependencies.
I can't figure out it's purpose

issmail-basel (dismissed):
I have 2 comments/suggestions to be checked.
Everything seems LGMT

src/components/organisms/Settings/SettingPage/hooks.ts
KaWaite: ohhh I'll double check but looks like maybe this was just a leftover from one of the useEffects I removed and didn't give me an error because it was still in the deps of the remaining one. Nice catch!

src/components/organisms/EarthEditor/Header/hooks.ts
KaWaite: I was thinking since team includes assets (which can be large) and other things having the Header update those when changing workspaces might not be great. I'll check if there are any issues with wrong assets, etc

issmail-basel: So inside `src/components/organisms/Dashboard/hooks.ts` file, is it better not to pass the whole object?
or there are there no assets in this object?

https://github.com/reearth/reearth-web/blob/414bf525a6240a7411cbcf09c1a10863d9d14e52/src/components/organisms/Dashboard/hooks.ts#L51-L54

PR #86, opened 2021-09-16
feat: add password validation

KaWaite:
# Overview
The passwordModal for changing your password merely disables the submit button unless new password and confirmation password match. This adds validation and feedback to make sure users are using secure passwords.

### DIFF
--- a/src/components/molecules/Settings/Account/PasswordModal/index.tsx
+++ b/src/components/molecules/Settings/Account/PasswordModal/index.tsx
@@ -6,7 +6,7 @@ import Flex from "@reearth/components/atoms/Flex";
 import Modal from "@reearth/components/atoms/Modal";
 import Text from "@reearth/components/atoms/Text";
 import TextBox from "@reearth/components/atoms/TextBox";
-import { styled } from "@reearth/theme";
+import { styled, useTheme } from "@reearth/theme";
 
 type Props = {
   className?: string;
@@ -28,7 +28,9 @@ type Props = {
 
 const PasswordModal: React.FC<Props> = ({ isVisible, onClose, hasPassword, updatePassword }) => {
   const intl = useIntl();
+  const theme = useTheme();
 
+  const [oldPassword, setOldPassword] = useState("");
   const [password, setPassword] = useState("");
   const [passwordConfirmation, setPasswordConfirmation] = useState("");
   const [disabled, setDisabled] = useState(true);
@@ -68,6 +70,7 @@ const PasswordModal: React.FC<Props> = ({ isVisible, onClose, hasPassword, updat
   }, [password, intl]); // eslint-disable-line react-hooks/exhaustive-deps
 
   const handleClose = useCallback(() => {
+    setOldPassword("");
     setPassword("");
     setPasswordConfirmation("");
     onClose?.();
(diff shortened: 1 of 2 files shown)


src/components/molecules/Settings/Account/PasswordModal/index.tsx
```diff
@@ -30,32 +31,97 @@ const PasswordModal: React.FC<Props> = ({ isVisible, onClose, hasPassword, updat
...
+  const [regexMessage, setRegexMessage] = useState("");
   const [passwordConfirmation, setPasswordConfirmation] = useState("");
   const [disabled, setDisabled] = useState(true);
 
+  const tooShortRegex = /^(?=.{1,7}$)/;
+  const tooLongRegex = /^(?=.{25,}$)/;
+  const whitespaceRegex = /(?=.*\s)/;
+  const highSecurityRegex = /^(?=.*[a-z])(?=.*[A-Z])((?=(.*\d){2}))/;
+  const medSecurityRegex = /^((?=.*[a-z])(?=.*[A-Z])|(?=.*[A-Z])(?=.*\d)|(?=.*[a-z])(?=.*\d))/;
+  const lowSecurityRegex = /^((?=\d)|(?=[a-z])|(?=[A-Z]))/;
```
rot1024, line 43: They are static values so they can be moved out of the component function, otherwise each regex will be re-created every render time

src/components/molecules/Settings/Account/PasswordModal/index.tsx
```diff
@@ -64,65 +130,71 @@ const PasswordModal: React.FC<Props> = ({ isVisible, onClose, hasPassword, updat
...
-            {intl.formatMessage({ defaultMessage: "New password (for confirmation)" })}
-          </Label>
-          <StyledTextBox
-            type="password"
-            borderColor={"#3f3d45"}
-            value={passwordConfirmation}
-            onChange={setPasswordConfirmation}
-          />
+            </Text>
+
```
rot1024, line 150: Does this blank line mean anything to you?

rot1024 (changes requested):

src/components/molecules/Settings/Account/PasswordModal/index.tsx
KaWaite: Nope!  I'll remove all unnecessary blanks!

rot1024: In HTML, blank lines are sometimes interpreted as whitespace; this doesn't happen in JSX, but it's confusing, isn't it?

KaWaite: haha ya I don't like random spaces either unless there is good reason.

src/components/molecules/Settings/Account/PasswordModal/index.tsx
```diff
@@ -64,65 +130,69 @@ const PasswordModal: React.FC<Props> = ({ isVisible, onClose, hasPassword, updat
...
+            {intl.formatMessage({
+              defaultMessage: `In order to protect your account, make sure your password:`,
+            })}
+          </Text>
+          <SubText>
+            <Text size="m">
               {intl.formatMessage({
-                defaultMessage: `In order to protect your account, make sure your password:`,
+                defaultMessage: `* Is between 8 and 25 characters in length`,
               })}
```
rot1024, line 148:
```jsx
<Text>
  *{" "}{intl.formatMessage({
    defaultMessage: `Is between 8 and 25 characters in length`,
  })}
</Text>
```

I think `*` is not a part of the text

rot1024 (commented):
It turns out that this PR actually contains a problem that we need to think about.

- Why do you limit passwords to between 5 and 25 characters? It is usually safe to not limit the maximum number of characters unless it is very long.
- Is the password policy in sync with the backend implementation and the OAuth provider's settings? If you don't have such a restriction in the backend or the provider, then trying to restrict it only in the frontend will result in a restriction that doesn't make much sense from a security perspective.
- Don't forget that users can also change the password from the login page provided by the OAuth provider. What if a different password policy is applied there? 

Re:Earth is OSS, so we never know who will use it and where will be used it, and even what OAuth2 provider will be used. Considering the possibility that various password policies may be required, it would be desirable to be able to change the password policy in, for example, reearth_config.json.

rot1024:
Additional implementation: to make password policy changeable by `reearth-config.json` (https://github.com/reearth/reearth-web/blob/main/src/config.ts) 

note: if no password policy is set, there is no limitation

PR #90, opened 2021-09-29
feat: enhance extended field of widget in plugin API

rot1024:
# Overview

## What I've done

```ts
type Widget {
  extended: boolean;
}
```

↓

```ts
type Widget {
  extended?: { horizontally: boolean; vertically: boolean; };
}
```

image

## How I tested

Manually

### DIFF
--- a/src/components/molecules/Visualizer/Widget/index.tsx
+++ b/src/components/molecules/Visualizer/Widget/index.tsx
@@ -1,10 +1,12 @@
 import React, { ComponentType, useMemo } from "react";
 
-import Plugin, { Widget, WidgetLayout } from "../Plugin";
+import Plugin, { Widget as RawWidget, WidgetLayout, WidgetLocation } from "../Plugin";
 
 import builtin from "./builtin";
 
-export type { Widget, WidgetLayout } from "../Plugin";
+export type { WidgetLayout } from "../Plugin";
+
+export type Widget = Omit<RawWidget, "layout" | "extended"> & { extended?: boolean };
 
 export type Props<PP = any, SP = any> = {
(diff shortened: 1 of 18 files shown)


HideBa (approved):

src/components/molecules/Visualizer/Widget/index.tsx
```diff
@@ -25,17 +31,25 @@ export default function WidgetComponent<PP = any, SP = any>({
...
+  const w = useMemo<RawWidget | undefined>(
     () =>
       widget
         ? {
             ...widget,
+            extended: !widgetLayout
+              ? undefined
+              : {
+                  horizontally: !!widget.extended && isHorizontal(widgetLayout.location),
+                  vertically: !!widget.extended && isVertical(widgetLayout.location),
```
KaWaite, line 43: Please correct me if I am wrong, but this code looks like it is setting horizontally or vertically to true depending on the Widget's current location (widgetLayout.location), right?
What we need the Widget component to know is if it (from the manifest) is **allowed** to be horizontal, vertical, both, or neither. Doesn't this code just **tell** the widget that it is allowed depending on its location and if it is extended?

KaWaite (commented):
Just one place I am not sure works the way it should.

src/components/molecules/Visualizer/Widget/index.tsx
rot1024: It is the backend's responsibility to restrict the extended, not the plugin's. Yes, and that information is enough for the plugin to set the style.

KaWaite (approved):

PR #115, opened 2021-10-28
feat: widget align system for mobile

KaWaite:
# Overview
Adding special mobile version of the widget align system.

## What I've done
- Add mobile version of the widget align system's sections based on the design by @lavalse 
- Update infobox (mobile AND desktop) so it is positioned better with the align system (it was still in the old position before the align system was implemented)

## What I haven't done
- Still some issues arise with the z-indexing between the widgets and the infobox. Needs further discussion and investigation

## Screenshot
Screen Shot 2021-11-02 at 18 32 52
Screen Shot 2021-11-02 at 18 33 21
Screen Shot 2021-11-02 at 18 33 31
Screen Shot 2021-11-02 at 18 37 18
Screen Shot 2021-11-02 at 18 37 28
Screen Shot 2021-11-02 at 18 41 33

## Things to test
- Infobox opens and minimizes (on layer selection, should be opened by default now)
- widget align system sections should be split in mobile
  - sections without widgets shouldn't be shown
  - align system with only one section populated with widgets should not show the bottom navigation
  - should start in middle section
  - should show arrow buttons if section in that direction is available.
  - should be able to select specific sections by clicking the middle circular buttons

### DIFF
--- a/.github/cliff.toml
+++ b/.github/cliff.toml
@@ -17,7 +17,7 @@ body = """
 {% for group, commits in commits | group_by(attribute="group") %}
     ### {{ group }}
     {% for commit in commits %}
-        - {{ commit.message | upper_first }} `{{ commit.id | split(pat="") | slice(end=7) | join(sep="") }}`\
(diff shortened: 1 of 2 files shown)


rot1024 (approved):
Excellent!

HideBa (approved):
LGTM

PR #127, opened 2021-11-09
fix: project setting page does not display correctly after creating a new project

issmail-basel:
# Overview
Project setting page does not display correctly after creating a new project

**To Reproduce**

1. Create a new project
2. Jump to the setting page of the project  newly created (do not show editor page)
3. Project name and left menus of the project doesn't display (reloading fixes it)

## What I've done
I've used the same project query in dashboard page as in settings page and same which is used in (refetch), so now refetch works fine as expected.

## How I tested
1. Create a new project
2. Jump to the setting page of the project  newly created (do not show editor page)
3. Project name and left menus of the project displays fine

## Screenshot

https://user-images.githubusercontent.com/78056580/141055279-52bc8cb4-b1d4-46ea-9305-c0670168a8ff.mp4

t


rot1024 (approved):
Videos are helpful to understand the situation. Thank you!

KaWaite (approved):

PR #130, opened 2021-11-15
feat: infinite scroll on assets

KaWaite:
# Overview

## What I've done
- Add infinite scroll to assets (Settings page and Asset Modals)
- Remove search and filter logic from front end and use query instead
- Assets in settings only have List and Medium card views, assets in modal have List and Small card views only
- Refactor
  - Consolidate asset related code into assetHooks
  - Remove as much duplicate code as possible (some places it wasn't possible, but with assetHooks its as good as can be)

## What I haven't done
- Dataset or project list infinite scroll (will be done in another PR as this one is already big). They shouldn't take long though since they don't have sort or search.

## Screenshot
Screen Shot 2022-03-08 at 12 18 46

## Memo
Had to switch away from relayStylePagination (which sucks) because there is an issue (as noted [here](https://github.com/apollographql/apollo-client/issues/8524) with RSP and hiding pagination related params in an object.


src/components/molecules/Settings/Workspace/Asset/AssetSection/index.tsx
```diff
@@ -12,23 +17,39 @@ type Asset = {
...
+    assets: AssetType[];
+    isLoading: boolean;
+    getMoreAssets: () => void;
+    createAssets: (files: FileList) => Promise<void>;
+    hasMoreAssets: boolean | undefined;
+    removeAsset: (assetIds: string[]) => Promise<void>;
+    sort?: { type?: AssetSortType | null; reverse?: boolean };
+    handleSortChange: (type?: string, reverse?: boolean) => void;
+    searchTerm?: string;
+    handleSearchTerm: (term?: string) => void;
```
rot1024, line 30: I like to put event props in the back of the type definition of props

```suggestion
    hasMoreAssets: boolean | undefined;
    sort?: { type?: AssetSortType | null; reverse?: boolean };
    searchTerm?: string;
    onRemoveAsset: (assetIds: string[]) => void;
    onSortChange: (type?: string, reverse?: boolean) => void;
    onSearchTerm: (term?: string) => void;
```

src/components/organisms/Common/AssetModal/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+} from "@reearth/gql";
+import { useNotification } from "@reearth/state";
+
+export type AssetNodes = NonNullable<AssetsQuery["assets"]["nodes"][number]>[];
+
+export enum AssetSortType {
+  Date = "DATE",
+  Name = "NAME",
+  Size = "SIZE",
+}
```
rot1024, line 19: It should be already defined in `@reearth/gql`

```suggestion
```

src/components/organisms/Common/AssetModal/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      return {
+        after: !reverseOrder ? endCursor : undefined,
+        before: reverseOrder ? endCursor : undefined,
+        first: !reverseOrder ? assetsPerPage : undefined,
+        last: reverseOrder ? assetsPerPage : undefined,
+      };
+    },
+    [sort],
+  );
```
rot1024, line 43: Rather than using useCallback, it would be more straightforward to make it a simple function and place it outside the hook.

```ts
function pagination(sort: { type?: Maybe<AssetSortType>; reverse?: boolean } | undefined, endCursor?: string | null) {
  const reverseOrder =
    (!sort?.type && !sort?.reverse) ||
    (sort?.type === "DATE" && !sort?.reverse) ||
    (sort?.type !== "DATE" && sort?.reverse);

  return {
    after: !reverseOrder ? endCursor : undefined,
    before: reverseOrder ? endCursor : undefined,
    first: !reverseOrder ? assetsPerPage : undefined,
    last: reverseOrder ? assetsPerPage : undefined,
  };
}
```

src/components/organisms/Common/AssetModal/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const { data, refetch, loading, fetchMore, networkStatus } = useAssetsQuery({
```
rot1024, line 45: If you close an asset modal and then reopen it, does it fetch data from the server again? If not, the cache has been saved and you should review the fetchPolicy.

src/components/organisms/Common/AssetModal/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  onSelect?: (asset: string | null) => void;
+}
+
+const AssetModal: React.FC<Props> = ({ teamId, openAssets, setOpenAssets, onSelect }) => {
+  const { assetsData } = useHooks(teamId);
+
+  return (
+    <MoleculeAssetModal
+      isOpen={openAssets}
+      onClose={() => setOpenAssets?.(false)}
```
rot1024, line 21: use useCallback to prevent extra render

src/gql/provider.tsx
```diff
@@ -55,9 +56,43 @@ const Provider: React.FC = ({ children }) => {
           },
         },
       },
+      Query: {
+        fields: {
+          assets: {
+            keyArgs: ["sort", "pagination", ["first", "last"]],
```
rot1024, line 62: relayStylePagination was not useful?

Is it because the backend implementation does not follow the relayStylePagination? I think it is better to keep the implementation as close as possible to standards and commonly accepted concepts. By using commonly accepted concepts, you reduce the learning cost for someone seeing the code for the first time.

Of course, this does not apply if there is a reasonable cause.

rot1024 (changes requested):

src/gql/provider.tsx
KaWaite: I wrote the reason in the PR description.

src/components/organisms/Common/AssetModal/hooks.ts
KaWaite: Should I delete the cache on close? Is there any reason we want a fresh AssetModal each time instead of leaving it as it was until the user changes sort, reverses or searches a word?
Isn't it more performant to not needlessly fetch if they have already been fetched?

src/components/organisms/Common/AssetModal/hooks.ts
KaWaite: This uses useState values inside the hooks. If it was a more generic function that could be used elsewhere I could understand why pulling it out like that would be useful.
Can you explain how it makes it simpler (since the kind of confusing conditional logic needs to remain)?

src/components/molecules/Settings/Workspace/Asset/AssetSection/index.tsx
```diff
@@ -12,23 +17,39 @@ type Asset = {
 };
 
 type Props = {
-  assets?: Asset[];
-  onCreate?: (files: FileList) => void;
-  onRemove?: (assetIds: string[]) => void;
+  assetsData: {
```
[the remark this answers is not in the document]
KaWaite: Up until now we don't wrap props, but is there a benefit to not wrap props into an object? 
For me, it makes reading the props more readable and a little easier to pass props.

src/components/molecules/Common/AssetModal/AssetContainer/hooks.ts
```diff
@@ -1,7 +1,7 @@
-import { useState, useCallback, useEffect } from "react";
+import { useState, useCallback } from "react";
 import useFileInput from "use-file-input";
 
-export type FilterTypes = "time" | "size" | "name";
+export type SortType = "DATE" | "NAME" | "SIZE";
```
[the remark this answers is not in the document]
KaWaite: Oh, @Basel-Issmail did similar for the Theme and that was merged so I kind of copied that style for consistency. Should creating convert functions for each case be the standard?

src/components/molecules/Settings/Workspace/Asset/AssetSection/index.tsx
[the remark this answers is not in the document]
rot1024: Flattening properties is more readable in JSX and more in line with React's style. The writing effort is no different thanks to the spread syntax (`<C {...props} />`).

src/components/organisms/Common/AssetModal/hooks.ts
rot1024:
```
function pagination(sortType: AssetSortType | undefined, reverse: boolean | undefined, endCursor?: string | null) {
  const reverseOrder = (!sortType && reverse) || (sortType === "DATE" ? !reverse : reverse);

  return {
    after: !reverseOrder ? endCursor : undefined,
    before: reverseOrder ? endCursor : undefined,
    first: !reverseOrder ? assetsPerPage : undefined,
    last: reverseOrder ? assetsPerPage : undefined,
  };
}
```

src/components/organisms/Common/AssetModal/hooks.ts
rot1024: The assumption is that the asset should not remain cached when the dialog is closed. The more paging is done, the more memory is consumed, and once the dialog is closed, the asset is no longer used, so the cache should be deleted when the component is unmounted.

And, since sorting or filtering of assets is done in AssetModal, there should be no concern as long as this component is mounted, the cache will persist even if the fetchPolicy is network-only. If the component is destroyed, the cache will be destroyed.

Please let me know if this perception is incorrect and actually works differently.

src/gql/provider.tsx
rot1024: Sorry I missed. OK I see.

src/components/molecules/Common/AssetModal/AssetContainer/hooks.ts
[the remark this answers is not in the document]
rot1024: I don't think there is anything to call a standard, but the world of GraphQL (or some other external technology) and the world of components should be loosely coupled, and if an implementor connects the two even unintentionally (for example, by sharing types between them), the one may be directly affected by the other. As one way to cleanly separate both worlds, I prefer to explicitly define a conversion function between the inner and outer worlds. It is also a matter of course if it is not defined, but then the implementer needs to be aware of it.

Example: GQL' s value is converted to reearth-web's value

https://github.com/reearth/reearth-web/blob/main/src/util/value.ts#L122

src/components/organisms/Common/AssetModal/hooks.ts
KaWaite: I figured out I can import useApolloClient and then, accessing the cache through there, evict the fieldName "assets" on close!

src/components/molecules/Settings/Workspace/Asset/AssetSection/index.tsx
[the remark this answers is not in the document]
KaWaite: I am updating to reflect your suggestion, but just for the record I am unable to use the spread syntax while also adhering to your other comment about keeping with react naming conventions of props (in the hooks the function `handleSortChange` would be passed to the component without being renamed based on the React convention onXXX (ie. onSortChange)). 
I'm assuming that is problematic and inconsistent so I'll write out each prop individually.

src/components/molecules/EarthEditor/PropertyPane/PropertyField/URLField/index.tsx
```diff
@@ -37,13 +36,29 @@ const URLField: React.FC<Props> = ({
...
+  assetSort,
+  assetSearchTerm,
+  onGetMoreAssets,
+  onCreateAssets,
+  onAssetSort,
+  onAssetSearch,
   onChange,
 }) => {
   const intl = useIntl();
+  const gqlCache = useApolloClient().cache;
```
rot1024, line 51: Rule Violation: molecules cannot use apollo directly!! `onAssetModalClose` and impl logic to delete cache in organisms

src/util/value.ts
```diff
@@ -80,6 +80,20 @@ const valueTypeMapper: Partial<Record<GQLValueType, ValueType>> = {
...
 export type ValueType = keyof ValueTypes;
 
+const enumTypeMapper: Partial<Record<any, string>> = {
+  [Theme.Default]: "default",
+  [Theme.Dark]: "dark",
+  [Theme.Light]: "light",
+  [AssetSortType.Date]: "date",
+  [AssetSortType.Name]: "name",
+  [AssetSortType.Size]: "size",
+};
```
rot1024, line 90: 1. Split (theme and asset type are entirely separate concerns)

```suggestion
const enumTheme = {
  [Theme.Default]: "default",
  [Theme.Dark]: "dark",
  [Theme.Light]: "light",
}

const enumAsserType = {
  [AssetSortType.Date]: "date",
  [AssetSortType.Name]: "name",
  [AssetSortType.Size]: "size",
};
```

2. Transfer them to each organisms hooks, because `src/util/value.ts` is the place to implement conversion between only "Property Values" ("value" in the file name means "property values").

src/util/value.ts
```diff
@@ -80,6 +80,20 @@ const valueTypeMapper: Partial<Record<GQLValueType, ValueType>> = {
...
+
+export const toGQLEnum = (val?: any) => {
+  if (!val) return;
+  return (Object.keys(enumTypeMapper) as any[]).find(k => enumTypeMapper[k] === val);
+};
+
```
rot1024, line 96: This function uses any, which may increase the number of bugs not found by compiler type checking. Implement the conversion function for each enum rather than loosening the type.

rot1024 (changes requested):

src/components/molecules/EarthEditor/PropertyPane/PropertyField/URLField/index.tsx
KaWaite: Since AssetContainer is a molecule itself, I wasn't able to move it down into it. I guess it is better to make an AssetContainer organism and move the cache deletion from AssetModal to there. If I do that, I won't have to pass onAssetModalClose prop as you suggested here!

src/util/value.ts
KaWaite: Oh okay. I didn't think it was really DRY to make a conversion function for each enum (for now there is only two, but later we could have more, multiplying the code). 
I guess, thinking about the rest of our typescript code, that it makes sense strong types > DRY code. 
I'll update now!

src/util/value.ts
KaWaite: Gotcha, makes sense. Thanks

src/components/molecules/EarthEditor/PropertyPane/PropertyField/URLField/index.tsx
KaWaite: Sorry, @rot1024 I remember now why I didn't do this in the first place. If I do as you ask here, I would need to create an AssetContainer organism, add the cache deletion there, and then the only way I can use it here in URLField is to pass the organism as a prop all the way down to here. 
The problem is as you can see, I am using local variables that I am unable to reach until the URLField. In particular `value`.
In saying that, would you say moving cache deletion is impossible? If so, your `onAssetModalClose` option seems the only option

rot1024: OrganismsAssetModal can accept selected value. Organisms components are not prohibited to accept normal props.

`src/components/organisms/Common/AssetModal/index.tsx`:

```
const OrganismsAssetModal = ({ value }: { value?: string }) => {
  // gql calls
  return <AssetMolecules ... />
}
```

KaWaite: Yes, but I'll need to pass `OrganismAssetModal` as a prop from (I believe) the `PropertyPane` organism, which means I can't pass it `value` from `URLField`, and to get the `value` in the `PropertyPane` organism would be close to impossible without a lot of work. Am I incorrect in this?

rot1024: How about changing props of molecule PropertyPanel?

```
type Props = {
 assetModal?: (value: string | undefined) => ReactNode
}
```

KaWaite: I was REALLY hoping something like that was possible, but I thought it was impossible. Thank you!

src/components/molecules/Common/AssetModal/AssetContainer/hooks.ts
```diff
@@ -16,30 +17,39 @@ export type Asset = {
...
   isMultipleSelectable?: boolean;
+  isLoading?: boolean;
   accept?: string;
   selectedAssets?: Asset[];
   sort?: { type?: SortType | null; reverse?: boolean };
   smallCardOnly?: boolean;
-  onCreateAsset?: (files: FileList) => void;
-  selectAsset?: (assets: Asset[]) => void;
+  onCreateAssets?: (files: FileList) => void;
+  selectAssetUrl?: (asset?: string) => void;
```
rot1024, line 38: It's an event prop

```suggestion
  onAssetUrlSelect?: (asset?: string) => void;
```

src/components/molecules/Common/AssetModal/AssetContainer/hooks.ts
```diff
@@ -16,30 +17,39 @@ export type Asset = {
...
 }) => {
+  const intl = useIntl();
   const [layoutType, setLayoutType] = useState<LayoutTypes>(smallCardOnly ? "small" : "medium");
   const [deleteModalVisible, setDeleteModalVisible] = useState(false);
 
+  const sortOptions: { key: SortType; label: string }[] = [
+    { key: "date", label: intl.formatMessage({ defaultMessage: "Date" }) },
+    { key: "size", label: intl.formatMessage({ defaultMessage: "File size" }) },
+    { key: "name", label: intl.formatMessage({ defaultMessage: "Alphabetical" }) },
+  ];
```
rot1024, line 51:
```suggestion
  const sortOptions: { key: SortType; label: string }[] = useMemo(() => ([
    { key: "date", label: intl.formatMessage({ defaultMessage: "Date" }) },
    { key: "size", label: intl.formatMessage({ defaultMessage: "File size" }) },
    { key: "name", label: intl.formatMessage({ defaultMessage: "Alphabetical" }) },
  ]), [intl]);
```

src/components/molecules/Common/AssetModal/AssetContainer/hooks.ts
```diff
@@ -93,12 +95,25 @@ export default ({
...
+    { currentTarget }: React.UIEvent<HTMLDivElement, UIEvent>,
+    onLoadMore?: () => void,
+  ) => {
+    if (
+      currentTarget.scrollTop + currentTarget.clientHeight >= currentTarget.scrollHeight &&
+      !isLoading
+    ) {
+      onLoadMore?.();
+    }
+  };
```
rot1024, line 108: This function can be put outside the hook

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -22,27 +22,35 @@ export type Asset = AssetType;
...
+  onCreateAssets?: (files: FileList) => void;
   onRemove?: (assetIds: string[]) => void;
   onGetMore?: () => void;
-  selectAsset?: (assets: Asset[]) => void;
+  selectAssetUrl?: (asset?: string) => void;
+  onSelect?: (asset?: Asset) => void;
   onSortChange?: (type?: string, reverse?: boolean) => void;
   onSearch?: (term?: string) => void;
-};
+  handleShowURL?: (assets?: Asset[]) => void;
```
rot1024, line 52:
```suggestion
  onURLShow?: (assets?: Asset[]) => void;
```

handle~ can be used for only local variables.

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -22,27 +22,35 @@ export type Asset = AssetType;
 
 export type AssetSortType = SortType;
 
+export type OtherProps = {
+  teamId?: string;
+  initialAssetUrl?: string | null;
+  allowDeletion?: boolean;
+};
+
```
rot1024, line 30: OtherProps seems not to be imported separatedly, so you can combien them into one `type Props = {...}`

src/components/molecules/Settings/Account/AccountSection/index.tsx
```diff
@@ -74,6 +74,14 @@ const ProfileSection: React.FC<Props> = ({
     [updatePassword],
   );
 
+  const handleThemeUpdate = useCallback(
+    (theme?: string) => {
+      if (!theme) return;
+      updateTheme?.(theme as Theme);
```
rot1024, line 80: Areas where bugs are likely to occur: Please avoid using "as" if at all possible, as is just "as" dangerous as any.

src/components/organisms/Settings/Account/hooks.ts
```diff
@@ -2,9 +2,21 @@ import { useApolloClient } from "@apollo/client";
...
 import { useTeam, useProject, useNotification } from "@reearth/state";
-import { toGQLEnum } from "@reearth/util/value";
+
+const enumTypeMapper: Partial<Record<GQLTheme, string>> = {
+  [GQLTheme.Default]: "default",
+  [GQLTheme.Dark]: "dark",
+  [GQLTheme.Light]: "light",
+};
+
+export type Theme = "dark" | "light";
```
rot1024, line 14: The types are not consistent. enumTypeMapper contains "default", but Theme type does not. it would be better to add "default" to Theme type.

```suggestion
export type Theme = "dark" | "light" | "default";
```

src/components/molecules/EarthEditor/PropertyPane/PropertyField/URLField/index.tsx
```diff
@@ -1,33 +1,18 @@
...
+import { Props as AssetModalPropsType } from "@reearth/components/molecules/Common/AssetModal";
 import { styled } from "@reearth/theme";
 
 import TextField from "../TextField";
 import { FieldProps } from "../types";
 
-export type Asset = AssetType;
-
-export type AssetSortType = SortType;
+export type AssetModalProps = AssetModalPropsType;
```
rot1024, line 11: For a more rigorous typing, you can write as following:

```suggestion
export type AssetModalProps = Pick<AssetModalPropsType, "isOpen" | "fileType" | "initialAssetUrl" | "onSelect" | "toggleAssetModal">;
```

rot1024 (changes requested):

src/components/molecules/Settings/Account/AccountSection/index.tsx
KaWaite: I would have done that, but the handleThemeUpdate is being used and not the updateTheme because it is being passed into the `EditableItem` component, which expects its `onSubmit` to be `(body: string) => void`.

KaWaite: Do you want me to update the EditableItem in this PR? I'm not great at generics so it might take a little while for me to sort it out. Also, we use that component in many sections of the Settings.

HideBa (approved):
LGTM 👍

rot1024 (approved):

PR #134, opened 2021-11-17
chore: upgrade dependencies

issmail-basel:
- Cypress was not updated.
Attempting to add an existing built-in Cypress command using Cypress.Commands.add() will now throw an error, indicating that Cypress.Commands.overwrite() should be used instead to overwrite the behavior of existing commands.
- Eslint emits error on a new rule `no-unsafe-optional-chaining` (other than that everything is identical)
I made a code change for it, please review
- @apollo/client was not updated
Gives exception `type 'ApolloLink' is not assignable to type 'ApolloLink | RequestHandler'` which does not have a clear solution


rot1024 (commented):
Great. Is tarn format done?

HideBa (approved):

rot1024 (approved):

PR #143, opened 2021-12-14
feat: clustering layers

issmail-basel:
# Overview

## What I've done
- Added clusters right Panel UI
- Added Cluster component
- Linked clusters front-end to back-end
- Add options to customize clusters

## How I tested
- Add new cluster
- Customize it's look
- Add layers to it
- Publish project

## Screenshot
Screenshot from 2021-12-17 05-46-36


rot1024 (approved):
Refactoring points (outside the scope of the PR):

- Standardize the implementation of Primitive component rendering: Cluster can accept children prop and render it.
- propertyRef in Cluster/index.tsx is not needed. The hooks can be simplified.

PR #149, opened 2021-12-21
refactor: camera limiter

issmail-basel:
# Overview

## What I've done
- Separate **camera limiter** logic into it's own file.
- Create clear pure functions for limiter calculations.

## What I haven't done
- I did not change the logic or any functionality code.

## How I tested
- same as the old camera limiter
- [x]  Editor
    - [x]  Choose a place and click on (camera input)
    - [x]  Show/hide helper
    - [x]  Enable/disable camera limiter
    - [x]  Change width/length of the area
    - [x]  Re-enable camera limiter
    - [x]  Preview result with limiter enabled and helper is on
        - [x]  Try to move to the edge
        - [x]  Try to zoom out
    - [x]  Preview result with limiter enabled and helper is off
    - [x]  Publish project with limiter enabled and helper is on
    - [x]  Publish project with limiter enabled and helper is off

## Screenshot
Screenshot from 2022-01-06 08-20-52

- Code enhancements/improvement.

### DIFF
--- a/src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
+++ b/src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
@@ -1,0 +1,239 @@
+import {
+  Cartesian3,
+  Cartographic,
+  Color,
+  EllipsoidGeodesic,
+  PolylineDashMaterialProperty,
+  Math,
+  Camera as CesiumCamera,
+  Rectangle,
+} from "cesium";
+import type { Viewer as CesiumViewer } from "cesium";
+import { useEffect, useMemo, useState } from "react";
+import { CesiumComponentRef } from "resium";
+import { RefObject } from "use-callback-ref/dist/es5/types";
+
+import { Camera } from "@reearth/util/value";
+
+import type { SceneProperty } from "..";
+
+import { getCamera } from "./common";
+
+export function useCameraLimiter(
+  cesium: RefObject<CesiumComponentRef<CesiumViewer>>,
+  camera: Camera | undefined,
+  property: SceneProperty["cameraLimiter"] | undefined,
+) {
+  const geodsic = useMemo(():
+    | undefined
+    | { geodesicVertical: EllipsoidGeodesic; geodesicHorizontal: EllipsoidGeodesic } => {
+    const viewer = cesium.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed()) return undefined;
+    return getGeodsic(viewer, property);
+  }, [cesium, property]);
+
+  // calculate inner limiter dimensions
+  const targetWidth = 1000000;
+  const targetLength = 1000000;
+  const limiterDimensions = useMemo((): InnerLimiterDimensions | undefined => {
+    const viewer = cesium.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed() || !geodsic) return undefined;
+    const {
+      cameraLimiterTargetWidth: width = targetWidth,
+      cameraLimiterTargetLength: length = targetLength,
+    } = property ?? {};
+
+    const { cartesianArray, cartographicDimensions } = calcBoundaryBox(
+      geodsic,
+      length / 2,
+      width / 2,
+    );
+
+    return {
+      cartographicDimensions,
+      cartesianArray,
+    };
+  }, [cesium, property, geodsic]);
+
+  // calculate maximum camera view (outer boundaries)
+  const [cameraViewOuterBoundaries, setCameraViewOuterBoundaries] = useState<
+    Cartesian3[] | undefined
+  >();
+
+  useEffect(() => {
+    const viewer = cesium.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed() || !property?.cameraLimiterTargetArea || !geodsic) return;
+
+    const camera = new CesiumCamera(viewer.scene);
+    camera.setView({
+      destination: Cartesian3.fromDegrees(
+        property.cameraLimiterTargetArea.lng,
+        property.cameraLimiterTargetArea.lat,
+        property.cameraLimiterTargetArea.height,
+      ),
+      orientation: {
+        heading: property?.cameraLimiterTargetArea.heading,
+        pitch: property?.cameraLimiterTargetArea.pitch,
+        roll: property?.cameraLimiterTargetArea.roll,
+        up: camera.up,
+      },
+    });
+
+    const computedViewRectangle = camera.computeViewRectangle();
+    if (!computedViewRectangle) return;
+    const rectangleHalfWidth = Rectangle.computeWidth(computedViewRectangle) * Math.PI * 1000000;
+    const rectangleHalfHeight = Rectangle.computeHeight(computedViewRectangle) * Math.PI * 1000000;
+
+    const {
+      cameraLimiterTargetWidth: width = targetWidth,
+      cameraLimiterTargetLength: length = targetLength,
+    } = property ?? {};
+
+    const { cartesianArray } = calcBoundaryBox(
+      geodsic,
+      length / 2 + rectangleHalfHeight,
+      width / 2 + rectangleHalfWidth,
+    );
+
+    setCameraViewOuterBoundaries(cartesianArray);
+  }, [cesium, property, geodsic]);
+
+  // Manage camera limiter conditions
+  useEffect(() => {
+    console.log("why");
+
+    const camera = getCamera(cesium?.current?.cesiumElement);
+    const viewer = cesium?.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed() || !property?.cameraLimiterEnabled || !limiterDimensions)
+      return;
+    if (camera) {
+      viewer.camera.setView({
+        destination: getAllowedCameraDestination(camera, limiterDimensions),
+        orientation: {
+          heading: viewer.camera.heading,
+          pitch: viewer.camera.pitch,
+          roll: viewer.camera.roll,
+          up: viewer.camera.up,
+        },
+      });
+    }
+  }, [camera, cesium, property, limiterDimensions]);
+
+  return {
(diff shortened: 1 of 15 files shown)


src/components/molecules/Visualizer/Engine/index.tsx
```diff
@@ -34,8 +34,8 @@ export type SceneProperty = {
     cameraLimitterEnabled?: boolean;
     cameraLimitterShowHelper?: boolean;
     cameraLimitterTargetArea?: Camera;
-    cameraLimitterTargetWidth: number;
-    cameraLimitterTargetLength: number;
+    cameraLimitterTargetWidth?: number;
+    cameraLimitterTargetLength?: number;
```
KaWaite, line 38: cameraLimiter, not cameraLimitter ;) One too many Ts

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+    | undefined
+    | { geodesicVertical: EllipsoidGeodesic; geodesicHorizontal: EllipsoidGeodesic } => {
+    const viewer = cesium.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed()) return undefined;
+    return getGeodsic(viewer, property);
+  }, [cesium, property]);
+
+  // calculate inner limiter dimensions
+  const targetWidth = 1000000;
+  const targetLength = 1000000;
```
KaWaite, line 37: Probably something we need to discuss a bit, but I really feel like these variables and the corresponding user-facing properties should be renamed since target really isn't meaningful. With that change, I feel like we can get rid of the explanatory comment as well (I don't feel like this place is complex or generic enough to need comments. Just a problem with variable naming).
Maybe we can discuss a bit before merging if we have time.

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+  cartographicDimensions: {
+    rightDemention: Cartographic;
+    leftDemention: Cartographic;
+    topDemention: Cartographic;
+    bottomDemention: Cartographic;
+  };
+  cartesianArray: Cartesian3[];
+};
+
+export const getGeodsic = (
```
KaWaite, line 141:
```suggestion
export const getGeodesic = (
```
!! Don't forget to update wherever this function is being called. !!

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+  const geodesicHorizontal = new EllipsoidGeodesic(
+    cartographicCenterPoint,
+    Cartographic.fromCartesian(east),
+    ellipsoid,
+  );
+  return { geodesicVertical, geodesicHorizontal };
+};
+
+export const calcBoundaryBox = (
+  geodsic: { geodesicVertical: EllipsoidGeodesic; geodesicHorizontal: EllipsoidGeodesic },
```
KaWaite, line 179:
```suggestion
  geodesic: { geodesicVertical: EllipsoidGeodesic; geodesicHorizontal: EllipsoidGeodesic },
```
!! Don't forget to update wherever this function is being called. !!

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+import type { SceneProperty } from "..";
+
+import { getCamera } from "./common";
+
+export function useCameraLimiter(
+  cesium: RefObject<CesiumComponentRef<CesiumViewer>>,
+  camera: Camera | undefined,
+  property: SceneProperty["cameraLimiter"] | undefined,
+) {
+  const geodsic = useMemo(():
```
KaWaite, line 27:
```suggestion
  const geodesic = useMemo(():
```
!! Don't forget to update wherever this variable is being used. !!

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+    leftDemention: Cartographic;
+    topDemention: Cartographic;
+    bottomDemention: Cartographic;
+  };
+  cartesianArray: Cartesian3[];
+} => {
+  const topDemention = geodsic.geodesicVertical.interpolateUsingSurfaceDistance(halfLength);
+  const bottomDemention = geodsic.geodesicVertical.interpolateUsingSurfaceDistance(-halfLength);
+  const rightDemention = geodsic.geodesicHorizontal.interpolateUsingSurfaceDistance(halfWidth);
+  const leftDemention = geodsic.geodesicHorizontal.interpolateUsingSurfaceDistance(-halfWidth);
```
KaWaite, line 194: xxDimension, not xxDemention

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+export const calcBoundaryBox = (
+  geodsic: { geodesicVertical: EllipsoidGeodesic; geodesicHorizontal: EllipsoidGeodesic },
+  halfLength: number,
+  halfWidth: number,
+): {
+  cartographicDimensions: {
+    rightDemention: Cartographic;
+    leftDemention: Cartographic;
+    topDemention: Cartographic;
+    bottomDemention: Cartographic;
```
KaWaite, line 187: xxDimension, not xxDemention

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+
+export function useCameraLimiter(
+  cesium: RefObject<CesiumComponentRef<CesiumViewer>>,
+  camera: Camera | undefined,
+  property: SceneProperty["cameraLimiter"] | undefined,
+) {
+  const geodsic = useMemo(():
+    | undefined
+    | { geodesicVertical: EllipsoidGeodesic; geodesicHorizontal: EllipsoidGeodesic } => {
+    const viewer = cesium.current?.cesiumElement;
```
KaWaite, line 30: Correct me if I am wrong, but it looks like every function and useEffect inside `useCameraLimiter` declares `viewer` and then returns undefined if there is no viewer or if the viewer is destroyed. 
Can't you just do it once in the useCameraLimiter's scope at the beginning? Then, you could simplify the other checks!

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,237 @@
...
+    if (!viewer || viewer.isDestroyed()) return undefined;
```
KaWaite, line 31: Correct me if I am wrong, but it looks like every function and useEffect inside useCameraLimiter declares viewer and then returns undefined if there is no viewer or if the viewer is destroyed.
Can't you just do it once in the useCameraLimiter's scope at the beginning? Then, you could simplify the other checks!

KaWaite (commented):
Please check my comments and update!

src/components/molecules/Visualizer/Engine/index.tsx
issmail-basel: Can't be updated unfortunately :( , as back-end manifest is already has the id of `cameraLimitter` in some places.
I have discussed it with Baba-chan and there is no way to update this until we have a way to migrate data (so camera limiter won't break for users who already use it).

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
issmail-basel: I think this can be done in the future release of Camera limiter (and I'll open an issue for it as well), as there are some UI changes as well in future.
For now the focus will be on code style.

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
issmail-basel: It would give me `tslint` errors, and I'll have to always use the viewer like this: `viewer?.xproperty` and Cesium functions require viewer to exist.
But there is one place where I don't need this check and I removed it from there :+1:

rot1024 (approved):

rot1024 (approved):

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
```diff
@@ -0,0 +1,241 @@
...
+        )
+      : undefined;
+  }, [cesium, property?.cameraLimitterTargetArea?.lng, property?.cameraLimitterTargetArea?.lat]);
+
+  // calculate inner limiter dimensions
+  const limiterDimensions = useMemo((): InnerLimiterDimensions | undefined => {
+    if (!geodesic) return undefined;
+
+    const width = property?.cameraLimitterTargetWidth || targetWidth;
+    const length = property?.cameraLimitterTargetLength || targetLength;
```
rot1024, line 48:
```suggestion
    const width = typeof property?.cameraLimitterTargetWidth === "number" ?  property.cameraLimitterTargetWidth : targetWidth;
    const length = typeof property?.cameraLimitterTargetLength === "number" ? property.cameraLimitterTargetLength : targetLength;
```

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/Cesium/cameraLimiter.ts
issmail-basel: @rot1024 , I thought there is no need for this, as the min value defined in manifest is `5` so this case won't happen.

rot1024 (approved):

PR #161, opened 2022-01-21
fix: cluster styling issue

issmail-basel:
# Overview
- If `Pixel range` is changed, cluster styling breaks
- `Minimum cluster size` is not working as expected.
- `UseEffect` is in a infinite loop.

## What I've done
- Used [cesium styling workaround](https://sandcastle.cesium.com/index.html?src=Clustering.html), without creating event loop 
- Added a workaround for when  `Minimum cluster size` is larger than number of cluster layers, without breaking the event listener.

## What I haven't done
- Did not add any features.

## How I tested
- Create a cluster.
- Change `pixel range`, cluster should not break.
- Choose `Minimum cluster size` to be larger than number of cluster layers.

## Screenshot
Before:
bug
After:
bfix


src/components/molecules/Visualizer/Engine/Cesium/Cluster.tsx
```diff
@@ -25,51 +25,63 @@ const Cluster: React.FC<ClusterProps> = ({ property, children }) => {
...
       pixelRange: clusterPixelRange,
-      minimumClusterSize: clusterMinSize,
+      minimumClusterSize: 2,
       clusterBillboards: true,
       clusterLabels: true,
       clusterPoints: true,
-    });
-  }, [clusterMinSize, clusterPixelRange]);
+    }),
+  );
```
rot1024, line 37: it looks like the useRef is being used incorrectly; there is no need to use it.

The performance is also worse because new EntityCluster is executed every time even though useRef is used. If this is the case, implement it without useRef first, and then consider how you can optimize the performance.

```ts
const cluster = useRef(
    new EntityCluster({
      // ...
    })
);
```

=>

```ts
  const culuster = new EntityCluster({
      enabled: true,
      pixelRange: clusterPixelRange,
      minimumClusterSize: clusterMinSize,
      minimumClusterSize: 2,
      clusterBillboards: true,
      clusterLabels: true,
      clusterPoints: true,
    });
```

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/Cesium/Cluster.tsx
issmail-basel: @rot1024 , 
Can I use `useMemo` with empty array (so it should compute once?) like this:

```
const cluster = useMemo(() => {
    return new EntityCluster({
      enabled: true,
      pixelRange: 15,
      minimumClusterSize: 2,
      clusterBillboards: true,
      clusterLabels: true,
      clusterPoints: true,
    });
  }, []);
```
or any suggestions.

src/components/molecules/Visualizer/Engine/Cesium/Cluster.tsx
```diff
@@ -25,51 +25,63 @@ const Cluster: React.FC<ClusterProps> = ({ property, children }) => {
...
-
-      // force a re-cluster with the new styling
-      cluster.pixelRange = 0;
-      cluster.pixelRange = clusterPixelRange;
-    });
+    const isClusterHidden = !!(
+      children &&
+      Array.isArray(children) &&
+      children.length < clusterMinSize
+    );
```
rot1024, line 44:
```suggestion
    const isClusterHidden = React.Children.count(children) < clusterMinSize;
```

src/components/molecules/Visualizer/Engine/Cesium/Cluster.tsx
rot1024: OK give it a shot.

rot1024 (approved):

HideBa (approved):

PR #163, opened 2022-01-24
fix: the style of infobox block dropdown list is broken

issmail-basel:
# Overview

## What I've done
- Removed `height` property from the parent.


rot1024 (changes requested):
What happens if 100 blocks are displayed? It may be extreme, but it is good to think that far. I would prefer to see scrollbars in the display for more than 4 lines (`max-height: XXpx; overflow: auto;`).

KaWaite (approved):

rot1024 (approved):

PR #170, opened 2022-02-04
feat: allowing widget and block plugins to resize when they are expandable

issmail-basel:
# Overview
Plugin developers want to be able to stretch 100% vertically or horizontally, and resize in the opposite direction. [related issue](https://github.com/reearth/reearth/issues/217)

## What I've done
- Auto resize now differentiate between two modes `autoResizeHorizontally` and `autoResizeVertically`.

## How I tested
- Added multiple widgets which supports **horizontal** and **vertical** extending.

## Screenshot
Before:

After:


src/components/atoms/Plugin/IFrame/hooks.ts
```diff
@@ -137,8 +137,17 @@ export default function useHook({
...
-        width: visible ? (autoResize ? iFrameSize?.[0] : "100%") : "0px",
-        height: visible ? (autoResize ? iFrameSize?.[1] : "100%") : "0px",
+        // TODO: width iFrameSize?.[0]
+        width: visible
+          ? autoResize == "width-only" || autoResize == "both"
+            ? "100%"
+            : iFrameSize?.[0]
+          : "0px",
+        height: visible
+          ? autoResize == "height-only" || autoResize == "both"
```
rot1024, line 147: resizable = variable (not fixed at 100%)
not resizable = fixed at 100%

```suggestion
          ? autoResize == "height-only" || autoResize == "both"
            ? "100%"
            : iFrameSize?.[0]
          : "0px",
        height: visible
          ? autoResize == "width-only" || autoResize == "both"
```

src/components/molecules/Visualizer/Plugin/hooks.ts
```diff
@@ -51,6 +51,13 @@ export default function ({
       ? `${pluginBaseUrl}/${`${pluginId}/${extensionId}`.replace(/\.\./g, "")}.js`
       : undefined;
 
+  const autoResize = useMemo((): "both" | "width-only" | "height-only" | undefined => {
+    if (!!widget?.extended?.horizontally && !!widget?.extended?.vertically) return "both";
+    if (widget?.extended?.horizontally) return "width-only";
+    if (widget?.extended?.vertically) return "height-only";
+    return undefined;
```
rot1024, line 58:
```suggestion
	const { horizontally, vertically } = widget?.extended ?? {};
    if (horizontally && vertically) return "both";
    if (vertically) return "width-only";
    if (horizontally) return "height-only";
    return undefined;
```

rot1024 (changes requested):

rot1024:
Change this also: https://github.com/reearth/reearth-web/blob/main/src/components/molecules/Visualizer/Block/index.tsx#L46

src/components/molecules/Visualizer/Plugin/hooks.ts
```diff
@@ -51,12 +53,14 @@ export default function ({
...
-    if (widget?.extended?.vertically) return "height-only";
+  const autoResizeDirection = useMemo((): "both" | "width-only" | "height-only" | undefined => {
+    if (autoResize) return autoResize;
+    const { horizontally, vertically } = widget?.extended ?? {};
+    if (horizontally && vertically) return "both";
+    if (vertically) return "width-only";
+    if (horizontally) return "height-only";
     return undefined;
-  }, [widget?.extended?.horizontally, widget?.extended?.vertically]);
+  }, [widget?.extended, autoResize]);
```
rot1024, line 63: Since the computation time is O(1) (= since there is no loop processing, the processing time is always constant, no matter what time it is.), the process is not as heavy as using useMemo. For such a simple process, the overhead of useMemo will be larger.

```ts
const actualAutoResize = autoResize ?? (
	widget?.extended?. horizontally && widget.extended.vertically ? "both" :
		widget?.extended?.vertically ? "width-only" :
			widget?.extended?. horizontally);
```

src/components/molecules/Visualizer/Plugin/hooks.ts
```diff
@@ -67,7 +71,7 @@ export default function ({
     onMessage,
     onPreInit,
     onDispose,
-    autoResize,
+    autoResizeDirection,
```
rot1024, line 74:
```suggestion
    actualAutoResize,
```

How about "actualAutoResize"?

rot1024 (approved):
I think it's good, but I have more in-depth advice.

PR #184, opened 2022-03-07
feat: basic plugin editor

KaWaite:
# Overview
A basic plugin editor inside of storybook to make developing Re:Earth plugins a little easier. Doesn't include YAML file editing, so you'll have to hardcode to see if editable fields will work or not, but it updates live so isn't all bad.

You can upload your local file and save a new one when you are done. You can see what it'll look like in different positions on the widget align system. You can test it as an infobox block and see it in small, medium or large infoboxes. And reset to your original if you need. Lastly, it has basic JavaScript syntax highlighting.

## Screenshot
Screen Shot 2022-03-18 at 17 18 02

## What I haven't done
Wishlist of what I want to add later:
- more features to the text editor (since it is using Monaco (what VS Code is built on) there is lots of possibilities to add to it.
- ability to extend your widget
- More templates (currently it is just "simple", with a single marker with an infobox. I'd like to add different templates for better testing (ie. template with many markers, photo overlays, tags, clusters, other widgets, etc, etc)


package.json
```diff
@@ -196,6 +198,7 @@
         "react-intl": "^5.24.6",
         "react-leaflet": "^3.2.5",
         "react-markdown": "^8.0.0",
+        "react-monaco-editor": "^0.47.0",
```
rot1024, line 201: I prefer https://github.com/suren-atoyan/monaco-react 

You may want to suspect that a library that uses class syntax in the README's React code is an unmaintained library.

package.json
```diff
@@ -180,6 +180,8 @@
         "leaflet": "^1.7.1",
         "lodash-es": "^4.17.20",
         "mini-svg-data-uri": "^1.4.3",
+        "monaco-editor": "^0.33.0",
+        "monaco-editor-webpack-plugin": "^7.0.1",
```
rot1024, line 184:
```suggestion
```

This library is not needed when you use monaco-react because monaco-editor on CDN will be used

https://github.com/suren-atoyan/monaco-react/issues/65#issuecomment-602836571

rot1024 (changes requested):
- Move `src/components/pluginEditor` to `src/components/molecules/PluginEditor` (rule violated)
- Reconsider the component name: I prefer `Playground` because the component is not truly an "editor" (we try plugins out rather than editing them)

package.json
KaWaite: Ok! I'll look into your suggested library.

rot1024 (changes requested):
Component directory name should be CamelCase:
   - ❌ `molecules/pluginEditor/...`
   - ✅ `molecules/PluginEditor/...`

About the name of PluginEditor:

I am still of the opinion that it cannot be called Editor; if it is going to call itself Editor, it needs to have the experience of being able to open a plugin directly on the file system, edit it, and save it on the fly. But the limitations of the web make that difficult. But even if you can't open and save a file, you can still try out the code. It is an experience just like [Go Playground](https://go.dev/play/), [TS Playground](https://www.typescriptlang.org/play), and [CodeSandbox](https://codesandbox.io/). 

This application will be placed as a page in the development documentation of the plugin. With that in mind, it sounds more logical that the plugin's "playground" would appear in the documentation rather than the plugin's "editor".

src/components/molecules/PluginEditor/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+          pluginBaseUrl={window.REEARTH_CONFIG?.plugins}
+          ready={true}
+          isEditable={true}
+          isBuilt={false}
+          small={false}
+          style={{ width: "100%" }}
+          widgets={{
+            ...(mode === "widget"
+              ? {
+                  alignSystem: alignSystem,
```
rot1024, line 49:
```suggestion
                  alignSystem,
```

src/components/molecules/PluginEditor/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+                }}>
+                <p>Extension type</p>
+                <select
+                  value={mode}
+                  onChange={e =>
+                    setMode(e.currentTarget.value as "block" | "widget" | "primitive")
+                  }>
+                  <option value="block">Block</option>
+                  <option value="widget">Widget</option>
+                  {/* <option value="primitive">Primitive</option> */}
```
rot1024, line 238:
```suggestion
```

There are currently no plans to support Primitive plug-ins

rot1024 (approved):
OK with some suggestions

PR #187, opened 2022-03-14
fix: display error messages from auth server

yk-eukarya:
# Overview


src/components/pages/Authentication/hooks.ts
```diff
@@ -100,20 +102,18 @@ export default () => {
       if (res.status !== 200) {
         setNotification({
           type: "error",
-          text: intl.formatMessage({ defaultMessage: "Something went wrong. Please try again." }),
+          text: res.data.error,
```
KaWaite, line 105: Is this error meaningful for the user? Most errors from the backend are not super meaningful and have zero punctuation (no periods and all lowercase). I know it is something eventually we want to update but for now unless we make a point to write user-friendly/user-facing error messages I feel like even a generic message like the one this change is replacing is better.
@rot1024 What do you think?

src/components/pages/Authentication/hooks.ts
```diff
@@ -1,24 +1,28 @@
...
   const passwordPolicy = window.REEARTH_CONFIG?.passwordPolicy;
 
   const { data, loading } = useTeamsQuery({ skip: !isAuthenticated });
   const teamId = currentTeam?.id || data?.me?.myTeam.id;
 
+  useEffect(() => {
+    if (location.pathname === "/login" && !new URLSearchParams(window.location.search).has("id"))
+      login();
+  }, [login, location.pathname]);
+
```
KaWaite, line 25: I believe I'm doing something similar in [this code block](https://github.com/reearth/reearth-web/blob/d824b61295603fd4348c29477ab0c8b88f799ed0/src/components/organisms/Authentication/RootPage/hooks.ts#L49-L62). Unless I am wrong and there is a reason for this, it might cause issues having both.

issmail-basel: The reason for this is when user navigate to `/login` without an id, it will re-generate the id, as requested recently on Notion.

src/components/pages/Authentication/hooks.ts
issmail-basel: @yk-eukarya , Yasser discuss this.

rot1024: Currently, you can either hide the error message at all, or you can just give the error message as it is.

Backends running in the production environment do not return error messages containing internal details. Error messages are written and human-readable, so it is acceptable to display them as they are. 

However, if this were to be made multilingual, additional implementation would be required. Since error messages are changed at the convenience of the backend, it would be better to include error message translations on the backend side, but so far no mechanism has been implemented for error messages based on the user's language preferences.

yk-eukarya: How about returning an error code with the message, something like this:
```
{
   meesage: "error message in English only",
   code: "E102"
}
```
the message can help the developer to figure out what is the error. the code can help to implement the translation on the front-end side. in this scenario, the translation has nothing to do with the back end and we can keep all translation stuffs will be in the front end.
@rot1024 What do you think?

rot1024: That method is vulnerable to change in our project structure. Every time we add or change the error messages in the backend, we have to change the frontend. It would be better if the backend had a translation of the error messages and returned the translated results.

rot1024 (approved):

PR #188, opened 2022-03-17
fix: duplicate asset results

KaWaite:
Fixes an issue where navigating to the asset's settings page would make it so the asset results would have duplicate items. 

There seems to be an issue with apollo for some users ([this issue ](https://github.com/vuejs/apollo/issues/185)or similar) where the merge callback would be called twice.

This solution uses lo-dash to check the equality of the existing cache and incoming query in the merge function, so that if they match, they won't be merged together.

### DIFF
--- a/src/components/organisms/Common/AssetContainer/hooks.ts
+++ b/src/components/organisms/Common/AssetContainer/hooks.ts
@@ -164,11 +164,9 @@ export default (teamId?: string, allowDeletion?: boolean) => {
 
   useEffect(() => {
     return () => {
-      setTimeout(() => {
-        setSort(undefined);
-        setSearchTerm(undefined);
-        gqlCache.evict({ fieldName: "assets" });
-      }, 200);
+      setSort(undefined);
+      setSearchTerm(undefined);
+      gqlCache.evict({ fieldName: "assets" });
     };
   }, [gqlCache]);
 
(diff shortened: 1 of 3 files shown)


src/components/organisms/Common/AssetContainer/hooks.ts
```diff
@@ -164,11 +164,9 @@ export default (teamId?: string, allowDeletion?: boolean) => {
...
   useEffect(() => {
     return () => {
-      setTimeout(() => {
-        setSort(undefined);
-        setSearchTerm(undefined);
-        gqlCache.evict({ fieldName: "assets" });
-      }, 200);
+      setSort(undefined);
+      setSearchTerm(undefined);
+      gqlCache.evict({ fieldName: "assets" });
```
HideBa, line 169: [imo] Since I'm not sure the reason you used `setTImeout` but if you still want to use it,

[must] you should make it like

```
useEffect(() => {
    const timer = 
      setTimeout(() => {
        setSort(undefined);
        setSearchTerm(undefined);
        gqlCache.evict({ fieldName: "assets" });
      }, 200);
      setSort(undefined);
      setSearchTerm(undefined);
      gqlCache.evict({ fieldName: "assets" });
	return clearTimeout(timer);
```

HideBa (commented):
I left a comment. Kindly check 👍

src/components/organisms/Common/AssetContainer/hooks.ts
rot1024: Does Baba's code clears cache on UNMOUNTING the component? I think it does not work..

rot1024 (approved):

src/components/organisms/Common/AssetContainer/hooks.ts
KaWaite: I tested it just in case, but no it doesn't. So I will not use this

HideBa (approved):

PR #192, opened 2022-03-25
fix: multiple assets in infinite scroll and datasets not showing in DatasetPane 

KaWaite:
# Overview
Infinite scroll on occasion will show multiples. When that happens, some items become unselectable as well.

Fixes https://github.com/reearth/reearth/issues/257
Fixes https://github.com/reearth/reearth/issues/256

## What I've done
- Fix bug in DatasetPane where datasets weren't showing
- Fixed the infinite scroll multiples bug
- Added unique message if no search results
- Regenerated translations (looks like we forgot to on auth functionality, etc)
- Fix some styling around the loading animation when loading assets

## How to test
- Try many different combinations of filtering and reversing. Quickly, slowly, half paginated, fully paginated, etc. (The bug is sometimes hard to reproduce so playing around and "trying to break it" is helpful)

## Memo
@rot1024 or @HideBa please add missing translations!


lavalse:
I have tested. I feel this PR solves the problem.  👍

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -159,8 +159,12 @@ const AssetContainer: React.FC<Props> = ({
       <AssetWrapper height={height}>
         {!isLoading && (!assets || assets.length < 1) ? (
           <Template align="center" justify="center">
-            <TemplateText size="m">
-              {fileType === "image"
+            <TemplateText size="m" otherProperties={{ textAlign: "center" }}>
+              {searchTerm
+                ? intl.formatMessage({
+                    defaultMessage: "No assets match your search.",
```
rot1024, line 165: 検索条件に該当するアセットは見つかりませんでした

PR #193, opened 2022-03-29
Fix the redirect after exc button in any setting page

nina992:
# Overview
When user presses Esc in any setting page, System redirects him to the setting page of workspace   

## What I've done
Add handling for ESC keypress in setting pages 

## How I tested
by going to all setting page and press esc then system redirects me to dashboard.

### DIFF
--- a/src/components/molecules/Settings/SettingPage/index.tsx
+++ b/src/components/molecules/Settings/SettingPage/index.tsx
@@ -1,5 +1,6 @@
-import { Link } from "@reach/router";
-import React, { useState } from "react";
+import { Link, navigate } from "@reach/router";
+import React, { useCallback, useState } from "react";
+import { useKeyPressEvent } from "react-use";
 
 import Icon from "@reearth/components/atoms/Icon";
 import Header, { Props } from "@reearth/components/molecules/Common/Header";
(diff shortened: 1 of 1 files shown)


src/components/molecules/Settings/SettingPage/index.tsx
```diff
@@ -18,7 +19,10 @@ const SettingPage: React.FC<Props> = ({
   const handleClick = () => {
     setIsOpen(o => !o);
   };
-
+  const handleKeyDown = useCallback(() => {
+    navigate(`/dashboard/${currentTeam?.id}`);
+  }, []);
```
KaWaite, line 24: Personally I don't feel like this redirect is desirable. I would be confused if I hit Escape in the Settings and got redirected to the dashboard. Was this change discussed with @lavalse (Red)? 

Ignoring that for a moment, my understanding of this bug is that somewhere in the code (or a package we use) we are redirecting on the escape key. 
I took a quick look around the Settings but couldn't see anything obvious, so it looks like you'll need to investigate a bit to figure out where the problematic redirect happens, and stop it. Unless @lavalse has a different idea, I feel like NOT having a redirect on ESC is better than having one.

KaWaite (changes requested):
Reviewed

KaWaite (approved):
Awesome works well!

src/components/atoms/Modal/ModalFrame/index.tsx
```diff
@@ -21,7 +21,9 @@ const Modal: React.FC<Props> = ({ size, isVisible, onClose, children }) => {
     unmountOnExit: true,
   });
 
-  const handleClose = useCallback(() => onClose?.(), [onClose]);
+  const handleClose = useCallback(() => {
+    isVisible && onClose?.();
```
rot1024, line 25:
```suggestion
    if (isVisible) onClose?.();
```

It is a matter of coding style preference, but I think the use of if statements would make the intent clearer here.

rot1024 (approved):

PR #195, opened 2022-03-30
fix: fix scrolling

nina992:
# Overview
the scrolling bar was inconsistence with the system style 
## What I've done
add scrollbar styles in global style file


KaWaite (changes requested):
Please move these into the app so we can utilize the theme context. Currently you are using background colors from the dark theme which won't look right if the user switches to light theme.
I messaged you on Slack as well, but let me know if you want me to explain more.

lavalse:
Because it works well on mac os. SO we close this PR for now

PR #198, opened 2022-04-08
fix: members section and new member button disappear after creating workspace from setting

nina992:
# Overview
When the user creates a new workspace from inside the setting page , the new workspace is created but the member section is empty and new member button disappears.

## What I've done
Change the graphql query and add the members to its result.


KaWaite (approved):
LGTM!

rot1024 (approved):

PR #201, opened 2022-04-14
fix: show unique icon for uploaded unsupported assets, plugin editor downloading only partial code, widget align system not wrapping in editor mode

KaWaite:
closes reearth/reearth#265
closes reearth/reearth#242

## What I've done
- Add unique icon for any assets that were uploaded before the recent asset modal infinite scroll and refactor/fix PR that are not officially supported.
- Fix renovate bot not running on schedule
- Fix plugin editor saving only partial value to downloaded file
- Fix WAS editor not wrapping components


src/components/molecules/Visualizer/WidgetAlignSystem/Area.tsx
```diff
@@ -57,6 +57,7 @@ export default function Area({
       align={(area === "middle" || section === "center") && widgets?.length ? align : undefined}
       style={{ flexWrap: wrapContent ? "wrap" : undefined }}
       editorStyle={{
+        flexWrap: wrapContent ? "wrap" : undefined,
```
issmail-basel, line 60: I prefer you put:
`flexWrap: wrapContent ? "wrap" : "nowrap",`
As it is clearer, and it is the flexWrap default value

src/components/molecules/PluginEditor/hooks.ts
```diff
@@ -69,6 +70,11 @@ export default () => {
     },
   ];
 
+  const handleFileDownload = (body?: string, filename?: string) => {
+    if (!body) return;
+    fileDownload(body, filename ?? "undefined.js");
```
issmail-basel, line 75: Why not making filename required?
What do you think?
If not, is `undefined.js` is the best name?

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
```diff
@@ -39,11 +39,7 @@ const AssetCard: React.FC<Props> = ({
       cardSize={cardSize}
       onClick={() => onCheck?.(!check)}>
       <ImgWrapper cardSize={cardSize}>
-        {url && /\.(jpg|jpeg|png|gif|svg|webp|GIF|JPG|JPEG|PNG|SVG|WEBP)$/.test(url) ? (
-          <PreviewImage url={url} />
-        ) : (
-          <Icon icon={icon} size={iconSize} />
-        )}
+        {!icon ? <PreviewImage url={url} /> : <Icon icon={icon} size={iconSize} />}
```
issmail-basel, line 42: I'm not sure, but what happens if someone tries to upload pdf file?
is there a place to handle this?

issmail-basel (commented):
Please check the comments, some of them are just questions to make sure that you handled all cases.

nina992 (approved):

src/components/molecules/Common/AssetModal/AssetCard/index.tsx
KaWaite: PDF files are not supported as Assets. Many files never were supported but we had not restricted uploading enough, or rather correctly. I removed this Regex test because, if you see in the AssetContainer, what kind of icon is chosen (or no icon for "image") is based on a fileType check. If you check the `checkFileType` function it does the checking on the supported types.
For AssetCard, it'll be no icon for images, file icon for supported file types, and then if the file type doesn't pass either checks, it will show an unsupported file icon.

src/components/molecules/PluginEditor/hooks.ts
KaWaite: Its not required because the variables we pass into this function come from `string | undefined` type. 
If you dislike the fallback name to be `undefined.js`, please suggest a better name and I'll update it.

PR #206, opened 2022-04-15
feat: implementation of the avatar feature in workspaces screens

nina992:
# Overview
A simple avatar feature makes the UX a little bit better, and also to solve some style issues like [Workspace banner stretching with many members ](https://github.com/reearth/reearth/issues/259)

 ## What I've done
1. Adding the avatar feature for below pages : 

- Workspace section in Dashboard page 
- Member list related to member section in workspace setting page
- Members section in each workspace in workspace list page
- profile picture in account page 

2. Match the new feature with the design (https://www.figma.com/file/bdnfDaGXGoNfUws4y7JItl/Re%3AEarth-UI-Gamma?node-id=9967%3A176976)

3. Main  changes in Code are : 
    1. Adding a new Atom component (Avatar)
    2. Adding a new color in both (light & dark  ) themes 
    3. Refactoring the MemberlistEdit component
    4. Adding  subHeader that could be used in field component

## What I haven't done
I tried to fix this bug [Broken alignment inside settings pages](https://github.com/reearth/reearth/issues/221)

## How I tested
1. Check the navigation bar avatar and make sure that it matches the design 
2. check the workspace section in dashboard and make sure it matches the design 
3. check the member list in workspace setting page make sure that the list shows the email- the right design
4. check the members in workspace list 

## Screenshot
image
 imageimage


src/components/molecules/Settings/Avatar/index.tsx
```diff
@@ -1,14 +1,17 @@
...
+  position: relative;
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
   min-width: ${({ size }) => size}px;
   min-height: ${({ size }) => size}px;
-  font-size: 24px;
-  border-radius: 50%;
-  background: ${({ avatar }) => (avatar ? `url(${avatar});` : "#888888")};
+  border-radius: ${({ radius }) => radius}%;
+  background: ${({ avatar, color }) => (avatar ? `url(${avatar});` : color)};
 `;
```
KaWaite, line 15: This should be made into a React component and moved into the atoms directory since it is simple and reused in multiple places.
Also, a couple additional things:
-  This should be useable on its own without more styling since in @lavalse's design the avatar looks the same (except for the `size`).  This means the Text component should be used inside this component (rather than how you coded it now, which you pass the Text component as a child everywhere this component is used).
- Personally I feel like this component should take a username and do the processing itself, rather than doing `avatar?.charAt(0).toUpperCase()` everywhere that this component is used.

This means the `Avatar` atom component's props should be: `size`, `username` and for future use `color`. A `radius` prop isn't needed bc we don't need a square or softened cornered square avatar.

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -11,6 +11,8 @@ import { Team as TeamType } from "@reearth/components/molecules/Dashboard/types"
 import { styled, useTheme, metrics } from "@reearth/theme";
 import { metricsSizes } from "@reearth/theme/metrics";
 
+import Avatar from "../Settings/Avatar";
```
KaWaite, line 14: We have TS and webpack setup to resolve the `@reearth/` alias to `src/`. As you can see, MOST imports are written from the `@reearth` alias, except for components that are one time use and nested (see [here](https://github.com/reearth/reearth-web/blob/cfcd605a6c319d2b6a0b29e75a6690f92e7be9b5/src/components/molecules/Common/AssetModal/AssetContainer/index.tsx#L15-L17) as an example), so please write most imports from `@reearth/`.

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -30,11 +33,20 @@ const Workspace: React.FC<Props> = ({ className, team }) => {
...
           <TeamWrapper flex={4}>
-            {team?.members?.map((member, i) => (
-              <Text key={i} size="m" color={theme.main.text}>
-                {member?.user.name}
-              </Text>
+            {team?.members?.slice(0, 5).map((member, i) => (
+              <StyledAvatar key={i} size={32} color={theme.main.avatarbg} radius={50}>
+                <Text size={isSmallWindow ? "m" : "l"} color={theme.text.pale}>
+                  {member?.user.name?.charAt(0).toUpperCase()}
+                </Text>
```
KaWaite, line 40: Please see Avatar component's comment

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -30,11 +33,20 @@ const Workspace: React.FC<Props> = ({ className, team }) => {
...
+              </StyledAvatar>
             ))}
+            {teamLength > 5 && (
+              <StyledAvatar size={32} color={theme.main.avatarbg} radius={50}>
+                <Text size={isSmallWindow ? "m" : "l"} color={theme.text.pale}>
+                  +{teamLength - 5}
+                </Text>
+              </StyledAvatar>
```
KaWaite, line 48: Please see Avatar component's comment

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -30,11 +33,20 @@ const Workspace: React.FC<Props> = ({ className, team }) => {
...
+            )}
```
KaWaite, line 49: Please see the comment on the `Avatar` component about changes to it.
For this code in particular:

This is fine, but I feel like if you use slice like this it isn't very clear why, or for what reason. 
If you can, one option would be instead of the slice here and declaring `teamLength` above, perhaps calculate `excessMembers`, and use that for filtering and conditionally rendering the `+` avatar.
Example:
```
const excessMembers = team?.members?.length - 5 ?? 0;

const shownMembers = useCallback(()=> ....return only first 5 members... );

return (
...
{shownMembers?.map((member, i) => (
   <Avatar key={i} value={member) />
)}
{excessMembers && (
 <Avatar value={excessMembers} /> 
)}
```

src/components/molecules/Settings/Avatar/index.tsx
KaWaite: If the Avatar components are USUALLY one size, and only sometimes another size, set a default size when destructuring your props in this component!

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -81,8 +93,12 @@ const StyledLink = styled(Link)`
 
 const TeamWrapper = styled(Flex)`
   * {
-    margin-right: ${metricsSizes.xl}px;
+    margin-right: ${metricsSizes.xs}px;
   }
+  row-gap: -5px;
+`;
+const StyledAvatar = styled(Avatar)`
```
KaWaite, line 100:
```suggestion
`;

const StyledAvatar = styled(Avatar)`
```

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -81,8 +93,12 @@ const StyledLink = styled(Link)`
 
 const TeamWrapper = styled(Flex)`
   * {
-    margin-right: ${metricsSizes.xl}px;
+    margin-right: ${metricsSizes.xs}px;
   }
+  row-gap: -5px;
```
KaWaite, line 98: Is this row-gap necessary if you are setting a margin to each avatar?

src/components/molecules/Settings/Project/EditableItem/index.tsx
```diff
@@ -24,6 +29,8 @@ export type Props<T extends string = string> = {
   currentItem?: T;
   imageSrc?: string;
   icon?: string;
+  isAvatar?: boolean;
+  avatar?: string;
```
KaWaite, line 33: `isAvatar` isn't necessary since you are passing `avatar` as well. Where you are using isAvatar, just do `!!avatar ? ... : ...` instead!

src/components/molecules/Settings/Project/EditableItem/index.tsx
```diff
@@ -158,3 +178,16 @@ const Image = styled.img`
...
+  justify-content: center;
+  align-items: center;
+  padding: 0px;
+  width: 64px;
+  height: 64px;
+  left: 150px;
+  top: 0px;
+  bottom: 2px;
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+`;
```
KaWaite, line 193: See the Avatar component's comment

src/components/molecules/Settings/Project/EditableItem/index.tsx
KaWaite: Just to note (and not something you need to do or should do in this PR): 

I think this component really needs a refactor. For now adding the avatar prop is fine. But ideally this component might be better with a React.componentType prop. That way we could pass an `<Icon />` or `<Avatar />` or something else and get rid of several props. 
Of course, there might be better ideas out there since this is just off the top of my head haha

src/components/molecules/Settings/Workspace/MemberListItem/index.tsx
```diff
@@ -1,17 +1,20 @@
...
+import { metricsSizes, styled, useTheme } from "@reearth/theme";
 
 import EditableItem from "../../Project/EditableItem";
 
 export type Role = "READER" | "WRITER" | "OWNER";
 
 export type Props = {
   name?: string;
+  email?: string;
   role: Role;
```
KaWaite, line 18: Instead of passing all three, can you condense this and just pass the whole user? The User does have id (which we won't use, and doesn't have role), but if you make it here:
```
user?: {
   name?: string;
   email?: string;
   role?: Role;
}
```
then I think you can pass it in as `user={...user, role}`.
What do you think?

src/components/molecules/Settings/Project/EditableItem/index.tsx
```diff
@@ -1,16 +1,21 @@
...
-import { styled } from "@reearth/theme";
+import { styled, useTheme } from "@reearth/theme";
 import { metricsSizes } from "@reearth/theme/metrics";
 
+import Avatar from "../../Avatar";
+
 export type Props<T extends string = string> = {
   className?: string;
   title?: React.ReactNode;
+  subTitle?: React.ReactNode;
```
KaWaite, line 18:
```suggestion
  subtitle?: React.ReactNode;
```

KaWaite: You'll need to update where this is used too.

src/theme/darkTheme.ts
```diff
@@ -25,6 +25,7 @@ const darkTheme: Theme = {
     link: colors.dark.functional.link,
     brandBlue: colors.brand.bg[1],
     brandRed: colors.brand.bg[2],
+    avatarbg: colors.dark.bg[5],
```
KaWaite, line 28:
```suggestion
    avatarBg: colors.dark.bg[5],
```

src/theme/lightheme.ts
```diff
@@ -25,6 +25,7 @@ const lightheme: Theme = {
     link: colors.light.functional.link,
     brandBlue: colors.brand.bg[1],
     brandRed: colors.brand.bg[2],
+    avatarbg: colors.light.bg[6],
```
KaWaite, line 28:
```suggestion
    avatarBg: colors.light.bg[6],
```

src/theme/theme.ts
```diff
@@ -38,6 +38,7 @@ export type Theme = {
     link: string;
     brandBlue: string;
     brandRed: string;
+    avatarbg: string;
```
KaWaite, line 41:
```suggestion
    avatarBg: string;
```

KaWaite (changes requested):
Made some suggestions, comments and asked for some changes!

Also, UI-based observations:

- Dashboard icon isn't centered
Screen Shot 2022-04-18 at 9 42 31
- Navbar avatar isn't updated
- The role and buttons (this screenshot doesn't show the buttons bc its my personal workspace) aren't centered really with the title AND subtitle

rot1024:
Rename the PR title because "avatar system" is abstract so that most people cannot understand what is changed in UI:

e.g. 

- feat: display avatars in the workspace screen

lavalse:
And also don't forget to update the navigation header 😂

lavalse:
And also update this page ( account setting page ) please .


design is [here](https://www.figma.com/file/bdnfDaGXGoNfUws4y7JItl/Re%3AEarth-UI-Gamma?node-id=6384%3A6795)

src/components/atoms/Avatar/index.stories.tsx
```diff
@@ -4,8 +4,8 @@ import React from "react";
 import Avatar from ".";
 
 export default {
-  title: "molecules/Settings/Avatar",
+  title: "atoms/Avatar",
   component: Avatar,
 } as Meta;
 
-export const Default = () => <Avatar size={24}>😄</Avatar>;
+export const Default = () => <Avatar size={24} />;
```
KaWaite, line 11: Since an avatar will have a letter, can you pass a userName for the storybook?

src/components/atoms/Avatar/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const Avatar: React.FC<Props> = ({ size = 24, avatar, color, userName, innerText }) => {
+  const theme = useTheme();
+  const isSmallWindow = useMedia("(max-width: 1024px)");
+  return (
+    <div>
+      <StyledAvatar size={size} avatar={avatar} color={color}>
+        {userName && (
+          <Text size={isSmallWindow ? "m" : "l"} color={theme.text.pale}>
+            {userName?.charAt(0).toUpperCase()}
```
KaWaite, line 24: @lavalse design doesn't look like the letter needs to be uppercase. Unless he wants to change it to be like that, you can remove the `toUpperCase()`

src/components/atoms/Avatar/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  avatar?: string;
+  color?: string;
+  userName?: string;
+  innerText?: string;
+};
+
+const Avatar: React.FC<Props> = ({ size = 24, avatar, color, userName, innerText }) => {
+  const theme = useTheme();
+  const isSmallWindow = useMedia("(max-width: 1024px)");
+  return (
```
KaWaite, line 19: This is maybe more of a preference, but if there is ANYTHING in the component other than the return I usually try to make a space between the return and the last variable or function.

src/components/atoms/Avatar/index.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
```
KaWaite, line 1: Let's pair pro this component today. There are a few changes I want to do and it'd be easier to pair pro than make comments for each!

KaWaite (approved):

src/components/atoms/Avatar/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { styled, useTheme } from "@reearth/theme";
+
+export type Size = "small" | "large";
+
+export type Props = {
+  className?: string;
+  size?: Size;
+  color?: string;
+  avatar?: string;
+  innerText?: string | number;
```
issmail-basel, line 14: The inner text should be string always, make the transform to string before passing it.
It's not meant to have `innerText` with a `number` type

src/components/atoms/Avatar/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <StyledAvatar
+      className={className}
+      align="center"
+      justify="center"
+      size={size}
+      avatar={avatar}
+      color={color ?? theme.main.avatarBg}>
+      {innerText && (
+        <Text size={"l"} color={theme.text.pale}>
+          {typeof innerText === "number" ? `+${innerText.toString()}` : innerText.charAt(0)}
```
issmail-basel, line 30: Can you explain this?

src/components/molecules/Dashboard/Workspace.tsx
```diff
@@ -21,6 +24,14 @@ const Workspace: React.FC<Props> = ({ className, team }) => {
   const theme = useTheme();
   const isSmallWindow = useMedia("(max-width: 1024px)");
 
+  const teamLength = team?.members?.length ?? 0;
+  const excessMembers = teamLength - 5 ?? 0;
+
+  const shownMembers: Member[] | undefined = useMemo(
```
issmail-basel, line 30: Is it agreed upon to show only the first 5?

src/components/molecules/Settings/Avatar/README.md
```diff
@@ -1,11 +0,0 @@
-# Account Settings Page
```
issmail-basel, line 1: Hmmm, why this file?

issmail-basel (changes requested):
Just a few comments to answer

issmail-basel (approved):

PR #208, opened 2022-04-19
refactor: queries/mutation code into a single directory

KaWaite:
# Overview
closes reearth/reearth#269

We had queries.ts files all over the place and unorganized (something like widget align system queries or mutations amongst layer related queries). To be more tidy, easier to update, search, and to maintain a standard that the gql fragments were already doing I moved all queries and mutations into a single directory.

Also, while testing, found that project thumbnails couldn't be updated anymore because of a backend api change, so fixed that bug as well.

## What I haven't done
- I haven't went through to see which queries and mutations are able to be combined/not used anymore so can be deleted/etc.
- I also haven't updated the names to have the same standard (some queries/mutations are  camel case, others are pascal case) since that would increase this PRs size astronomically. Also, might be a good task for a new frontend dev member.


rot1024 (changes requested):
1. Mixed plural and singular in file name, so it would be great to unify the file names to one of them. (e.g. `tags.ts` -> `tag.ts`)
2. "Primitive" is a word not used in GraphQL schema, so merge `primitive.ts` to `scene.ts`

rot1024 (approved):

PR #211, opened 2022-04-20
chore: add a code style guide

KaWaite:
# Overview
As our team grows, we have begun to realize we need a consolidated style guide to keep the code clean, easy to read, organized and not a burden to newcomers both in the Core team and the Community.

## What I've done
This is a preliminary start with intentions to add, update and reshape it as the team sees fit.

## What I haven't done
I haven't written an exhaustive guide that needs no further changes. To the best of my ability I have outlined some very common and important points to the current code base as well as some goal posts that will help make the source code more readable. More discussions are needed to further flesh this out.


rot1024 (changes requested):

CODE_STYLE.md
```diff
@@ -0,0 +1,268 @@
...
+
+Use to avoid unnecessary re-rendering of processed objects or variables, but only use it when computational cost is high (think **more than** `O(n)` in cost).
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```TypeScript
+const item = data?.find(d => d.id === id);
```
issmail-basel, line 92: Isn't the complexity/cost of the `find` function O(n)?

issmail-basel (approved):
I only have one comment

CODE_STYLE.md
KaWaite: Yep, that is why it is an example of bad, and that we should use useMemo in that situation instead! ;)

issmail-basel: @KaWaite , my mistake, I misunderstood.
I thought what you meant by "(think more than O(n) in cost)." means `>` O(n) not `>=` O(n) :smile:

KaWaite: Ah maybe I should reword it

CODE_STYLE.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+</td><td>
+
+```TypeScript
+const item = useMemo(() => {
+  return data?.find(d => d.id === id);
+}, [data, id]);
```
rot1024, line 98:
```suggestion
const item = useMemo(() => {
  return data?.find(d => d.id === id);
}, [data, id]);
const object = useMemo(() => { id }, [id]);
```

CODE_STYLE.md
```diff
@@ -0,0 +1,260 @@
...
+
+Use to avoid unnecessary re-rendering of processed objects or variables, but only use it when computational cost is high (think **more than** `O(n)` in cost).
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```TypeScript
+const item = data?.find(d => d.id === id);
```
rot1024, line 90:
```suggestion
const item = data?.find(d => d.id === id);
const object = { id };
```

CODE_STYLE.md
```diff
@@ -0,0 +1,260 @@
...
+*Note: Due to having eslint setup to catch quite a few syntax and code layout issues out of the box, we have not included those here. Please make sure your IDE is setup properly.*
+
+## Guidelines
+---
+
+### React
+
+**useMemo**
+
+Use to avoid unnecessary re-rendering of processed objects or variables, but only use it when computational cost is high (think **more than** `O(n)` in cost).
```
rot1024, line 82:
```suggestion
- Use to avoid unnecessary re-processing and re-rendering of processed objects or variables, but only use it when computational cost is high (think **more than** `O(n)` in cost).
- Use to avoid unnecessary re-rendering of props and status values wrapped in objects or arrays.
```

CODE_STYLE.md
```diff
@@ -0,0 +1,260 @@
...
+</td></tr>
+</tbody></table>
+
+### Spacing and order
+
+*NOTE: Often ESlint will catch spacing issues on save, but not always. Because of this, be mindful of your spacing as it affects the file's readability immensely. For order, ESlint will only fix import order.*
+
+Since file content varies in size, complexity, whether it is a component or utility, hooks, etc., it is hard to set specific rules for spacing and order. We only ask that you be deliberate (and take time refactoring before opening a PR) and be open to suggestions on spacing or order during PR review to keep code as readable and accessible to new eyes as possible.
+
+### Destructuring
```
rot1024, line 234: This rule can be detected by ESLint, so it would be better to set up ESLint than to write it here, so everyone will abide by this rule.

https://eslint.org/docs/rules/prefer-destructuring

CODE_STYLE.md
```diff
@@ -0,0 +1,260 @@
...
+
+**useCallback**
+
+Use for MOST functions within a React component to better handle renders based on dependencies. Even if no dependencies, a useCallback will make sure React knows the function will always be the same, so don’t need to re-render
+
+### GraphQL
+
+Query and mutation exports should live inside the `src/gql/queries` directory only.  *Contents* of each file should be queries, followed by mutations. 
+
+Naming of queries and mutations should be in PascalCase.
```
rot1024, line 112:
```suggestion
- Naming of variables in TypeScript codes should be in upper SNAKE_CASE.
- naming of queries and mutations should be in PascalCase.
```

rot1024 (changes requested):

CODE_STYLE.md
KaWaite: Ok, I'll look into that!

rot1024 (approved):

PR #212, opened 2022-04-26
chore: introduce i18next

rot1024:
# Overview

## What I've done

- Introduce i18nnext
- Add a npm script: `yarn i18n` (so fast!!!)
- Rename `gen:18n` script to `i18n:legacy`
- Rename `gen:gql` script to `gql` (fewer type times)

## What I haven't done

react-intl is not uninstalled yet. The number of translations is so large that it would be difficult to migrate them all at once, so we would like to rewrite them using i18next when we see react-intl when we revamp a component.

## How I tested

Check Netlify preview to confrim the app continues to work


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+    "editor.formatOnSave": false, // conflicts with ESLint format
```
KaWaite, line 2: If other devs need to update/revert their IDE settings, please let us know on slack!

i18next-parser.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  locales: ["en", "ja"],
+  output: "src/translations/$LOCALE.yml",
```
KaWaite, line 3: Is there any reason we will keep translations separated from i18n code? In CMS and Cloud we are storing everything together, so if there is a good reason to keep them separate I will update CMS and cloud to keep things streamlined. If not, please output to 
```suggestion
  output: "src/i18n/translations/$LOCALE.yml",
```

package.json
```diff
@@ -24,9 +24,9 @@
         "storybook:screenshot": "storycap --serverCmd 'npx http-server ./storybook-static -p 6006' --serverTimeout 100000 http://localhost:6006 --max-old-space-size=2048",
         "cypress:open": "cypress open",
         "cypress:run": "cypress run",
-        "gen": "run-p gen:*",
-        "gen:gql": "graphql-codegen --config codegen.yml",
-        "gen:i18n": "extract-messages -d en -l=en,ja -o translations -f yaml 'src/**/*.ts*' --extractFromFormatMessageCall",
+        "gql": "graphql-codegen --config codegen.yml",
+        "i18n": "i18next",
+        "i18n:regacy": "extract-messages -d en -l=en,ja -o src/translations/legacy -f yaml 'src/**/*.ts*' --extractFromFormatMessageCall",
```
KaWaite, line 29:
```suggestion
        "i18n:legacy": "extract-messages -d en -l=en,ja -o src/translations/legacy -f yaml 'src/**/*.ts*' --extractFromFormatMessageCall",
```

src/i18n/publishedProvider.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { PropsWithChildren } from "react";
+import { I18nextProvider } from "react-i18next";
+
+import i18n from "./i18n";
+import { PublishedProvider as LegacyProvider } from "./legacy";
+
+export default function PublishedProvider({ children }: PropsWithChildren<{}>) {
```
KaWaite, line 7: Based on [this post](https://stackoverflow.com/a/71809927) by one of React's devs, he says PropsWithChildren doesn't need to be used, and is only useful for the `codemod`. 
Also, children typings are fixed in React 18, so we will need to start providing a prop type. Can you add that here? Type for children: `children?: React.ReactNode`

src/translations/en.yml
```diff
@@ -0,0 +1,2 @@
+Continue: ''
+Cancel: ''
```
KaWaite, line 2: Is this correct? I haven't had a chance to look into i18next enough yet, but I was hoping we could setup the generation to be better than how it is in cloud (manual) and cms (I think same as here). Benefits of this current way is the Japanese will be easier to translate (bc the english can be seen in ja.yml), so maybe this way is ok? Just `EnglishWord: ''` seems a little strange to me.

src/i18n/provider.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useLanguageQuery } from "@reearth/gql";
+
+import i18n from "./i18n";
+import { Provider as LegacyProvider } from "./legacy";
+
+export default function Provider({ children }: { children?: ReactNode }) {
+  const { isAuthenticated } = useAuth();
+  const { data } = useLanguageQuery({ skip: !isAuthenticated });
+  const locale = data?.me?.lang;
+  useEffect(() => {
```
KaWaite, line 14:
```suggestion
  const locale = data?.me?.lang;
  
  useEffect(() => {
```

src/i18n/provider.tsx
```diff
@@ -0,0 +1,23 @@
...
+    i18n.changeLanguage(locale);
```
KaWaite, line 15: `locale`
Screen Shot 2022-04-27 at 11 42 20
, if not chosen by the user yet, apparently can be `und` which is used for the `Auto` setting.

KaWaite (commented):
Reviewed!

.vscode/settings.json
rot1024: OK, let me discuss about it later. For now, that line has been reverted.

src/translations/en.yml
rot1024: Setting useKeysAsDefaultValue to true will cause the key name to be inserted directly into the value.

src/i18n/provider.tsx
rot1024: I tried to set undefined as the language (i.e. automatic) when und. But it needs to be tested.

rot1024: tested and it works

KaWaite (approved):
Looks good!

PR #214, opened 2022-04-27
fix: color picker auto-complete with unexpected 'ff'

airslice:
closes reearth/reearth#237

# Overview

When user input a RGB in hex 6(eg. #ff9900) Re:earth will auto complete it with a 'ff' for alpha channel (eg.#ff9900ff) which makes a little confuse.
After discusstion the solution is setting the placeholder as '#RRGGBBAA' to indicate user there's two space for alpha channel.

## What I've done

Change the placeholder of color input from 'Not set' to '#RRGGBBAA'

## Screenshot

image


issmail-basel (approved):

airslice:
going to re-create branch and make a new pr

PR #220, opened 2022-05-09
feat: Add opacity field to map tiles

KaWaite:
# Overview
closes reearth/reearth#277
Users want to be able to layer map tiles and change the opacity to see the tile below the top one.

## What I've done
Added a SliderField component to the right panel. It should be ok for other fields that need a slider as well, since its marks, min, max, step are all customizable with a "reasonable" fallback.

## What I haven't done
The SliderField steals lots of its logic from the Slider atom. There might be a way to consolidate into one component, but at the moment they are so different looking that it didn't seem worth the time to try. Keeping them as their own entities.

## How I tested
Added multiple map tiles and played around with different opacities. 
Change using slider.
Change using input (both with number pad and going up by steps with arrow keys on blur.

## Screenshot
Screen Shot 2022-05-09 at 17 08 06
Screen Shot 2022-05-09 at 17 08 35


src/components/molecules/EarthEditor/PropertyPane/PropertyField/SliderField/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+import { FieldProps } from "../types";
+
+export type Props = FieldProps<number> & {
+  min?: number;
+  max?: number;
+  step?: number;
+};
+
+const inputRegex = new RegExp(/^\d+\.\d{2,}$/);
```
rot1024, line 16:
```suggestion
const inputRegex = /^\d+\.\d{2,}$/;
```

src/components/molecules/EarthEditor/PropertyPane/PropertyField/SliderField/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const isEditing = useRef(false);
+  const isDirty = useRef(false);
+
+  const calculatedStep = step ? step : max ? max / 10 : 0.1;
+
+  const opacityMarkers = {
+    [min ?? 0]: min ?? 0,
+    [max ? max / 2 : 0.5]: max ? max / 2 : 0.5,
+    [max ?? 1]: max ?? 1,
+  };
```
rot1024, line 39:
```suggestion
  const opacityMarkers = useMemo(() => ({
    [min ?? 0]: min ?? 0,
    [max ? max / 2 : 0.5]: max ? max / 2 : 0.5,
    [max ?? 1]: max ?? 1,
  }), [min, max]);
```

useMemo is just helpful here

src/components/molecules/EarthEditor/PropertyPane/PropertyField/SliderField/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+          onChange(floatValue);
+          isDirty.current = false;
+        }
+      }
+    },
+    [onChange, min, max, value],
+  );
+
+  const handleInputChange = useCallback((e?: React.ChangeEvent<HTMLInputElement>) => {
+    const target = e?.currentTarget;
```
rot1024, line 73:
```suggestion
  const handleInputChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
    const target = e.currentTarget;
```

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -51,24 +51,42 @@ export default ({
 
   // imagery layers
   const [imageryLayers, setImageryLayers] =
-    useState<[string, ImageryProvider, number | undefined, number | undefined][]>();
+    useState<
+      [string, ImageryProvider, number | undefined, number | undefined, number | undefined][]
```
rot1024, line 55: It's hard to understand the slice type when you get to this point. You can define a type:

```suggestion
      ImageryLayerData[]
```

```ts
type ImageryLayerData = {
  id: string;
  provider: ImageryProvider;
  min: number | undefined;
  max: number | undefined;
  opacity: number  | undefined;
}
```

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -51,24 +51,42 @@ export default ({
...
+        id,
+        type ? (url ? imagery[type](url) : imagery[type]()) : null,
+        min,
+        max,
+        opacity,
+      ])
       .filter(
-        (t): t is [string, ImageryProvider, number | undefined, number | undefined] => !!t[1],
+        (
+          t,
```
issmail-basel, line 82: @KaWaite  , can you explain this to me?

issmail-basel (approved):

rot1024 (approved):

PR #222, opened 2022-05-10
fix: Query names in refetchQueries not updated

airslice:
# Overview

The query names used in refetchQuery are not updated, which will lead to some components can't get auto updated after some mutaions.

## What I've done

Updated the query names in refetchQuery of hooks:
- GetAssets
- GetTeams
- GetSceneTags
- GetLayerTags
- GetProjects

## How I tested

Create & remove projects / teams(workspace) / scene tags / layer tags, upload / remove assets of image.


KaWaite (approved):
Thanks for doing this so quickly!

PR #223, opened 2022-05-11
chore: add useT hook to i18n

rot1024:
- I added `useT` hook to `@reearth/i18n` that enables writing translations with less typing.
   - NOTE: Do not use other variable names than `t`, as translation will not be visible to the i18n-parser
- Disable i18n-parser option `useKeysAsDefaultValue` to clarify whether each text is translated or not. There is no effect on the on-screen display (if empty, the key is used as is).

### DIFF
--- a/src/i18n/index.tsx
+++ b/src/i18n/index.tsx
@@ -1,3 +1,8 @@
+import { useTranslation } from "react-i18next";
+
 export { useTranslation } from "react-i18next";
 export { default as Provider } from "./provider";
 export { default as PublishedProvider } from "./publishedProvider";
+
+// eslint-disable-next-line react-hooks/rules-of-hooks
+export const useT = () => useTranslation().t;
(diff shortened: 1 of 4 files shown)


src/i18n/index.tsx
```diff
@@ -1,3 +1,8 @@
+import { useTranslation } from "react-i18next";
+
 export { useTranslation } from "react-i18next";
 export { default as Provider } from "./provider";
 export { default as PublishedProvider } from "./publishedProvider";
+
+// eslint-disable-next-line react-hooks/rules-of-hooks
+export const useT = () => useTranslation().t;
```
KaWaite, line 8: Is there a reason we need to keep this a function, and not just do `export const t = useTranslation().t;`? 
If we did that we could `import {t} from "@reearth/i18n";` and use t immediately, rather than always declaring useT, cleaning up code more.

KaWaite (approved):
Approved with a question about implementation. When you have time if you could answer that would be awesome!

src/i18n/index.tsx
rot1024: useTranslation is "a React hook" so it should be always called inside components.

PR #224, opened 2022-05-12
feat: Add scene property overriding to Re:Earth API

KaWaite:
# Overview
closes reearth/reearth#276

To extend the abilities and possibilities of plugin development, the scene property is able to be overridden by a plugin (ex. for Cesium, skybox can be triggered, atmospheric conditions can be changed, etc by plugins).

## What I've done
Added a `overrideProperty` method to ReEarth API. This allows plugins to target specific scene property fields and modify them ephemerally (temporarily).
Now do a merge to sceneProperty. If no overriddenSceneProperty, will return the project's sceneProperty. If something is overridden, it'll merge into the project's.

## How I tested
- Changed multiple scene property fields successfully from the api


src/components/molecules/Visualizer/hooks.ts
```diff
@@ -52,6 +62,25 @@ export default ({
 }) => {
   const engineRef = useRef<EngineRef>(null);
 
+  const [overriddenSceneProperty, overrideSceneProperty] = useState({});
+
+  const handleScenePropertyOverride = useCallback((property: any) => {
+    overrideSceneProperty(p => mergeProperty(p, property));
+  }, []);
```
rot1024, line 69:
```suggestion
  const handleScenePropertyOverride = useCallback((property: any) => {
    overrideSceneProperty(property);
  }, []);
```

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -52,6 +62,25 @@ export default ({
...
+      return sceneProperty;
+    return mergeWith(sceneProperty, overriddenSceneProperty, (obj: any, other: any) => {
+      if (Array.isArray(obj) && Array.isArray(other)) {
+        const merged = merge({}, keyBy(obj, "id"), keyBy(other, "id"));
+        return values(merged);
+      } else {
+        return merge({}, obj, other);
+      }
+    });
+  }, [sceneProperty, overriddenSceneProperty]);
```
rot1024, line 82:
```suggestion
  const mergedSceneProperty = useMemo(() => {
    return mergeProperty(sceneProperty, overriddenSceneProperty)
  }, [sceneProperty, overriddenSceneProperty]);
```

rot1024 (changes requested):
- Define `mergeProperty` function at `VIsualizer/util.ts` and then use it at `Visualzier/hooks.ts` and `Visualizer/Primitive/index.tsx`
- `overriddenSceneProperty` should just keep an object passed from plugins as is instead of the merged result to simplify implementation.

src/components/molecules/Visualizer/hooks.ts
KaWaite: That is how I originally had it. But that meant if, for example, I changed the tile type with one button, and then I changed the opacity with a slider widget, the changed tile type would revert back to the original. So only if the plugin remembers the changes, or does ALL changes at once, would multiple changes persist.
Do you have a better solution for that issue?

src/components/molecules/Visualizer/hooks.ts
KaWaite: I believe this works if I make the above suggested change, but doesn't taking in to consideration what I mention above. I'll wait for the above reply before making any changes here.

src/components/molecules/Visualizer/hooks.ts
rot1024: @KaWaite OK, It is better to keep the override property for each plugin and merge them in one batch.

```ts
const [overriddenSceneProperty, overrideSceneProperty] = useState({});
```

->

```ts
const [overriddenSceneProperty, overrideSceneProperty] = useState<{ [pluginId: string]: any }>({});
```

---

```ts
const handleScenePropertyOverride = useCallback((property: any) => {
  overrideSceneProperty(property);
}, []);
```

->

```ts
const handleScenePropertyOverride = useCallback((pluginId: string, property: any) => {
  overrideSceneProperty(p => (pluginId && property ? { ...p, [pluginId]: property } : omit(p, pluginId));
}, []);
```

omit is lodash function

---

```ts
const mergedSceneProperty = useMemo(() => {
    return mergeProperty(sceneProperty, overriddenSceneProperty)
  }, [sceneProperty, overriddenSceneProperty]);
```

->

```ts
const mergedSceneProperty = useMemo(() => {
    return Object.values(overriddenSceneProperty).reduce((p, v) => mergeProperty(p, v), sceneProperty);
  }, [sceneProperty, overriddenSceneProperty]);
```

src/components/molecules/Visualizer/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { cloneDeep, mergeWith } from "lodash";
+
+export function mergeProperty(a: any, b: any) {
+  const a2 = cloneDeep(a);
+  return mergeWith(
+    { ...a2 },
```
rot1024, line 6:
```suggestion
    a2,
```

Already cloned

rot1024 (approved):
Note that ci prefix cannot be used for this PR

PR #228, opened 2022-05-20
Feat：Add scene mode selector

airslice:
@close https://github.com/reearth/reearth/issues/283

# Overview

## What I've done

Add scene mode selector for switch modes between 3D/2D/2.5D(a.k.a. columbusView).
Basic function for scene switch.

## What I haven't done

Haven't handle the bugs running at 2D/2.5D mode. These will be handled in the later SBLs.

## How I tested

Select different modes from the selector and play with the earth.

## Screenshot

image
image
image


src/util/value.ts
```diff
@@ -24,6 +24,8 @@ export type Camera = {
   fov: number;
 };
 
+export type SceneMode = "scene3d" | "scene2d" | "columbus";
```
KaWaite, line 27: Either this is in the wrong spot, or you need to add this to the valueTypeMapper below as it asks. 
(Pretty sure the types here are for the typescript versions of GraphQL types and we use these + the mapper in the `valueFromGQL` function to convert. So if thats not how this type is being used, move it somewhere more appropriate.)

src/components/molecules/Visualizer/Engine/Cesium/useEngineRef.ts
```diff
@@ -84,6 +84,23 @@ export default function useEngineRef(
...
+          case "scene2d":
+            viewer?.scene?.morphTo2D(duration);
+            break;
+          case "columbus":
+            viewer?.scene?.morphToColumbusView(duration);
+            break;
+          default:
+            break;
+        }
+      },
```
KaWaite, line 103: Since `scene3d` is the default, can we write it like this:
```suggestion
        switch (sceneMode) {
          case "scene2d":
            viewer?.scene?.morphTo2D(duration);
            break;
          case "columbus":
            viewer?.scene?.morphToColumbusView(duration);
            break;
          case "scene3d":
          default:
            viewer?.scene?.morphTo3D(duration);
            break;
        }
      },
```
Unless there is a reason you want the default to not do any morphing??

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -130,6 +130,11 @@ export default ({
     [property?.default?.bgcolor],
   );
 
+  // scene mode
```
KaWaite, line 133: I don't think we need this comment since the code is pretty self-explanatory

src/components/molecules/Visualizer/Engine/Cesium/common.ts
```diff
@@ -287,31 +288,38 @@ export const animateFOV = ({
...
   if (!viewer || viewer.isDestroyed() || !viewer.camera || !viewer.scene) return undefined;
   const { camera } = viewer;
-  if (!(camera.frustum instanceof PerspectiveFrustum)) return;
-
-  const ellipsoid = viewer.scene.globe.ellipsoid;
-  const { latitude, longitude, height } = ellipsoid.cartesianToCartographic(camera.position);
+  if (
+    !(
+      camera.frustum instanceof PerspectiveFrustum ||
+      camera.frustum instanceof OrthographicOffCenterFrustum
```
KaWaite, line 314: Can you explain why we check if camera.frustum is an instanceof `OrthographicOffCenterFrustum` here? 
The PerspectiveFrustum makes sense, but not the OrthographicOffCenterFrustum. Does it make more sense for you to be checking if `camera.positionCartographic`, or am I missing something?

airslice: The `camera.frustum` can be some other types which are not in use for the system by now. I checked this for returning expect result only when it is  `PerspectiveFrustum` for 3d/2.5d mode and `OrthographicOffCenterFrustum` for 2d mode.

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
airslice: updated

src/util/value.ts
airslice: Sorry didn't notice this is for convert values. Moved into the `ref.ts` now.

KaWaite (approved):
LGTM

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -14,12 +14,15 @@ export type EngineRef = {
...
   zoomIn: (amount: number) => void;
   zoomOut: (amount: number) => void;
+  morphSceneModeTo: (sceneMode: SceneMode | undefined, duration?: number) => void;
   isMarshalable?: boolean | "json" | ((target: any) => boolean | "json");
   builtinPrimitives?: Record<string, Component>;
   pluginApi?: any;
   clusterComponent?: ComponentType<ClusterProps>;
 };
 
+export type SceneMode = "scene3d" | "scene2d" | "columbus";
```
rot1024, line 24:
```suggestion
export type SceneMode = "3d" | "2d" | "columbus";
```

simple name

src/components/molecules/Visualizer/Engine/Cesium/useEngineRef.ts
```diff
@@ -84,6 +84,22 @@ export default function useEngineRef(
         if (!viewer || viewer.isDestroyed()) return;
         viewer?.scene?.camera.zoomOut(amount);
       },
+      morphSceneModeTo: (sceneMode, duration = 2) => {
```
rot1024, line 87:
```suggestion
      changeSceneMode: (sceneMode, duration = 2) => {
```

rot1024 (approved):

HideBa (approved):
LGTM! 👍

PR #229, opened 2022-05-24
chore: update dependencies for reearth-web

nina992:
# Overview
This Pr is for upgrading project dependencies.

## What I've done
Run the command :  yarn upgrade-interactive --latest 

## How I tested
Run yarn test 
Run yarn type 
Test the system functionality


package.json
```diff
@@ -53,28 +53,32 @@
             "<rootDir>/node_modules/"
         ]
     },
+    "resolutions": {
+        "@types/react": "17.0.14",
+        "@types/react-dom": "17.0.14"
+    },
```
KaWaite, line 59: @nina992 Can you explain why these are necessary? I think they came up in a solution you found for a problem you were having in your local environment, but it would be better if you fixed your local environment rather than adding to our codebase (unless absolutely necessary).

PR #230, opened 2022-05-25
fix: iframe not correctly sizing to plugin

KaWaite:
closes reearth/reearth#275

The iframe for plugins was either too big or too small, not adhering to width and height set by the plugin itself.
This fixes that and also re-adds the plugin link to the project menu in the center of the header.

Extra: 
I think it is how the plugin API is designed, but just to reiterate: plugins need to set the width and height in both the html itself and the `reearth.ui.show(html, {height: xx, width: xx})` for this to work.


src/components/molecules/Common/ProjectMenu/index.tsx
```diff
@@ -46,6 +46,12 @@ const ProjectMenu: React.FC<Props> = ({ currentProject, teamId }) => {
                 text={intl.formatMessage({ defaultMessage: "Datasets" })}
               />
             </MenuListItem>
+            <MenuListItem>
+              <MenuListItemLabel
+                linkTo={`/settings/project/${currentProject.id}/plugins`}
+                text={intl.formatMessage({ defaultMessage: "Plugins" })}
```
rot1024, line 52: use i18next

rot1024 (approved):

PR #233, opened 2022-05-27
feat: adapt camera field to support 2d mode

airslice:
close https://github.com/reearth/reearth/issues/285

# Overview

This task is aiming to update CameraField and related (like FovSlider) at UI level to support mode 2d.
Basiclly hiding or disable the pose parameters / UI. 

## What I've done

- add a `onlyPosition` parameter to `CameraField`
- get the sceneMode value from gql and create a status and passing it down to the `CameraField`
- hide the pose part of `CameraField` when use 2d mode
- disable `CameraField` when `onlyPose` and `onlyPosition` at the same time
- hide the `FovSlider` when use 2d mode

## How I tested

Switch scene mode and test CameraField and FovSlider

## Screenshot

image

image

i meet the requirement that the property value inside the right panel (sceneProperty.default.sceneMode) will have effects on the other components (like CameraField for now) also inside the right panel (include the ones in other groups). The way i handle this is maybe not a good way i think.


src/components/organisms/EarthEditor/PropertyPane/hooks-queries.ts
```diff
@@ -191,6 +191,12 @@ export default ({
 
   const layers = useMemo(() => convertLayers(layerData), [layerData]);
 
+  const sceneMode = useMemo(() => {
+    const defaultPropery = sceneProperty?.items.find(item => item.schemaGroupId === "default");
```
KaWaite, line 195: defaultProperty

airslice:
updates:
- add a sceneMode status 
- add a setter inside the CanvasArea as it always get the scene property
- remove the property value finder for scene mode from the hooks-queryies of PropertyPane

src/components/organisms/EarthEditor/PropertyPane/hooks-queries.ts
airslice: this part has been removed.

src/state/index.ts
```diff
@@ -34,6 +34,10 @@ export const useIsCapturing = () => useAtom(isCapturing);
 const camera = atom<Camera | undefined>(undefined);
 export const useCamera = () => useAtom(camera);
 
+export type SceneMode = "3d" | "2d" | "columbus";
+const sceneMode = atom<SceneMode | undefined>("3d");
```
KaWaite, line 38: Is this actually ever undefined? It starts as "3d" and I'm guessing there isn't (or shouldn't be) any way to set it to undefined, right? Or is this based on Cesium types?

KaWaite (approved):
Just one comment. If it is necessary the way it is, all looks fine!

src/state/index.ts
airslice: emmm yes there shouldn't be any way to set it to undefined. i'll update this part and the setter.

PR #236, opened 2022-05-30
fix: vertical position style in infobox image block

keiya01:
# Overview

I fixed reearth/reearth#202.

## What I've done

I fixed to use `height: 100%;` when vertical position property is specified center in infobox image block.

## Screenshot

|Before|After|
|-|-|
|Screen Shot 2022-05-30 at 20 57 28|Screen Shot 2022-05-30 at 20 57 38|


KaWaite (approved):
Thanks!

rot1024 (approved):

PR #237, opened 2022-05-31
chore: update config so extensionUrls can be declared in .env file for local development

KaWaite:
# Overview
Just update config.ts so that devs can put an extension urls into their .env files locally

In .env
`REEARTH_WEB_EXTENSION_URLS=https://somewhere.com`
or multiple
`REEARTH_WEB_EXTENSION_URLS=https://somewhere.com,https://anothersomewhere.com,https://moresomewheres.com`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -62,7 +62,7 @@ export type Config = {
     highSecurity?: RegExp;
   };
   ip?: string;
-  extensionURLs?: string[];
+  extensionUrls?: string[] | string;
   extensions?: Extensions;
 };
 declare global {
@@ -156,8 +156,11 @@ export default async function loadConfig() {
     );
   }
 
-  if (window.REEARTH_CONFIG?.extensionURLs) {
-    const urls = window.REEARTH_CONFIG.extensionURLs;
+  if (window.REEARTH_CONFIG?.extensionUrls) {
+    let urls = window.REEARTH_CONFIG.extensionUrls;
+    if (!Array.isArray(urls)) {
+      urls = [urls];
+    }
     const extensions = getExtensionsFrom(urls);
     window.REEARTH_CONFIG.extensions = extensions;
   }
--- a/src/config.ts
+++ b/src/config.ts
@@ -62,7 +62,7 @@ export type Config = {
     highSecurity?: RegExp;
   };
   ip?: string;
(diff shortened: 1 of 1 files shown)


src/config.ts
```diff
@@ -156,8 +156,11 @@ export default async function loadConfig() {
...
   }
 
-  if (window.REEARTH_CONFIG?.extensionURLs) {
-    const urls = window.REEARTH_CONFIG.extensionURLs;
+  if (window.REEARTH_CONFIG?.extensionUrls) {
+    let urls = window.REEARTH_CONFIG.extensionUrls;
+    if (!Array.isArray(urls)) {
+      urls = urls.split(",");
+    }
     const extensions = getExtensionsFrom(urls);
```
rot1024, line 164:
```suggestion
    const extensions = getExtensionsFrom(window.REEARTH_CONFIG.extensionUrls);
```

URL can contain commas (,) ([RFC2396](https://www.ietf.org/rfc/rfc2396.txt)), so it is not best way to get an array. It is best to convert it to an array when loading .env (it can be implemented in webpack.config.js). When reearth_config.json is used, it is sufficient to load it as is.

rot1024 (changes requested):

src/config.ts
KaWaite: Ah....didn't know that.
I tried last night to find a way to do the converting in webpack.config.js, but we are using loadEnv to get the file (it is a Buffer), parsing the envfile, and then using readEnv to manipulate it more, and then spreading (...) itinto the config. I couldn't figure out how to do it it since I am unfamiliar with Buffers and got a little confused with everything that is happening there.
If you could give me an idea where the conversion should happen, maybe I can figure it out. Or if we want to do this quickly, maybe if you could give me the implementation I can update the PR

rot1024: Parsing array is already done! So just merge my suggestion, and then it works.

We are using https://github.com/yatki/read-env and it supports multiple values:

```
EXAMPLE_ARRAY='[1,2,3, "string", {"prop": "value"}, 5.2]'
```

can be converted to

```json
{
  "array": [1, 2, 3, "string", { "prop": "value" }, 5.2]
}
```

rot1024 (approved):

PR #238, opened 2022-06-02
fix: labeling hidden by marker symbol

nourbalaha:
# Overview
In this PR I fixed the bug where the label is hidden behind the marker image
this fixes reearth/reearth#207 and reearth/reearth#184 reearth-web

## What I've done
I modified the `pixelOffset` hook inside `src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx`
and updated the equations for x and y with taking imageSize(imageScale) in condsideration.

## How I tested
Manually using google chrome.

## Screenshot
image

### DIFF
--- a/src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
+++ b/src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
@@ -143,7 +143,7 @@ const Marker: React.FC<PrimitiveProps<Property>> = ({ layer }) => {
         ? y * (labelPos.includes("top") ? -1 : 1)
         : 0,
     );
-  }, [style, imgw, pointSize, imgh, labelPos]);
+  }, [isStyleImage, imgw, pointSize, imgh, labelPos]);
 
   const e = useRef<CesiumComponentRef<CesiumEntity>>(null);
   useEffect(() => {
--- a/src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
+++ b/src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
@@ -134,7 +134,7 @@ const Marker: React.FC<PrimitiveProps<Property>> = ({ layer }) => {
     const padding = imageSize ? 15 : 40;
     const x =
       (img?.width && imageSize && (!style || style === "image")
-        ? img?.width * imageSize
+        ? img.width * imageSize
         : pointSize) /
         2 +
       padding;
--- a/src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
+++ b/src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
@@ -114,8 +114,9 @@ const Marker: React.FC<PrimitiveProps<Property>> = ({ layer }) => {
       : undefined;
   }, [extrude, location, height]);
 
-  const [icon, img] = useIcon({
-    image,
+  const isStyleImage = !style || style === "image";
+  const [icon, imgw, imgh] = useIcon({
+    image: isStyleImage ? image : undefined,
     imageSize,
     crop,
     shadow,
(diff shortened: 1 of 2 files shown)


src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
```diff
@@ -131,9 +131,19 @@ const Marker: React.FC<PrimitiveProps<Property>> = ({ layer }) => {
   );
 
   const pixelOffset = useMemo(() => {
-    const padding = 15;
-    const x = (img?.width && style == "image" ? img.width : pointSize) / 2 + padding;
-    const y = (img?.height && style == "image" ? img.height : pointSize) / 2 + padding;
+    const padding = imageSize ? 15 : 40;
```
rot1024, line 134:
```suggestion
    const padding = 15;
```

What is your intention in changing the padding? If you don't have any intention, please change it back.

rot1024 (changes requested):
Thank you for your first PR!

src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
nourbalaha: The problem we are having right now is the following:
when we upload an image for a marker it will be very small and imageSize(imageScale) will be undefined until we adjust the imageSize(imageScale) manually, that is why I added a margin of 40 in case the imageSize(imageScale) is undefined to avoid overlapping in both cases.

nourbalaha: this is how the label and image look before manually setting the imageSize(imageScale)
image

src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
```diff
@@ -131,9 +131,19 @@ const Marker: React.FC<PrimitiveProps<Property>> = ({ layer }) => {
...
+        ? img.width * imageSize
+        : pointSize) /
+        2 +
+      padding;
+    const y =
+      (img?.height && imageSize && (!style || style === "image")
+        ? img.height * imageSize
+        : pointSize) /
+        2 +
+      padding;
```
rot1024, line 146:
```suggestion
    const x =
      (img?.width && (!style || style === "image")
        ? img.width * (imageSize ?? 1)
        : pointSize) /
        2 +
      padding;
    const y =
      (img?.height && (!style || style === "image")
        ? img.height * (imageSize ?? 1)
        : pointSize) /
        2 +
      padding;
```

Does this not work?

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
rot1024: I see. More suggestion: https://github.com/reearth/reearth-web/pull/238#discussion_r888564906

src/components/molecules/Visualizer/Engine/Cesium/Marker/index.tsx
nourbalaha: In this suggestion when we upload a new image the imageSize will be undefined by default that is why (imageSize ?? 1) will always be 1 at the beginning.
however when we upload the image, the size of the image is not relevant to the imageSize and all images will be the same size regarding of the imageSize

Screen Shot 0004-06-03 at 11 46 41 AM

and then when we change the imageSize manually it will adjust to the actual imageSize

Screen Shot 0004-06-03 at 11 47 44 AM

so in your suggestion we will have something like this at the start before adjusting the imageSize manually

Screen Shot 0004-06-03 at 11 52 27 AM

rot1024: I see. You mean `imageSize` will not be undefined thanks to `const { imageSize = 0.1 } = property?.default ?? {};`.

So in your expression,

```
const padding = imageSize ? 15 : 40;
```

I think `padding` should be always `15` unless `0` is intentionally specified for `imageSize`. Is it incorrect?

nourbalaha: Not exactly what I meant.
because I have not added `const{ imageSize = 0.1 } = property?.default ?? {};` to the code base yet.
what I mean is the padding will always be 15 unless image size is 0 or undefined(at the beginning).

rot1024 (approved):

PR #240, opened 2022-06-04
refactor: migrate to react-intl from react-i18next

keiya01:
# Overview

## What I've done

I replaced react-intl with react-i18next.
And I removed react-intl pkg.
closes https://github.com/reearth/reearth/issues/172.

## How I tested

1. Access to reearth staging environment
2. Check all of text

- I replaced useIntl with useT.
- I replaced ja translation file by using the following script.

<details>
<summary>replacing script</summary>
<pre><code>
import fs from "node:fs/promises";

import yaml from "js-yaml";

const findKeyByValueRecursiveObj = (obj, val) => {
  const keys = Object.keys(obj);
  for (const key of keys) {
    const objValue = obj[key];
    if (typeof objValue === "string") {
      if (objValue === val) return key;
      continue;
    }
    const foundKey = findKeyByValue(objValue, val);
    if (foundKey) {
      return foundKey;
    }
  }
};

const findValueByKeyRecursiveObj = (obj, key) => {
  const keys = Object.keys(obj);
  for (const objKey of keys) {
    const objValue = obj[objKey];
    if (typeof objValue === "string") {
      if (objKey === key) return objValue;
      continue;
    }
    const foundValue = findValueByKey(objValue, key);
    if (foundValue) {
      return foundValue;
    }
  }
};

const findKeyByValue = (obj, val) => {
  if (Array.isArray(obj)) {
    throw Error("Array is not supported");
  } else {
    return findKeyByValueRecursiveObj(obj, val);
  }
};

const findValueByKey = (obj, key) => {
  if (Array.isArray(obj)) {
    throw Error("Array is not supported");
  } else {
    return findValueByKeyRecursiveObj(obj, key);
  }
};

// You need to run this script after `yarn i18n`.
const run = async () => {
  const [legacyEnRaw, legacyJaRaw, newJaRaw] = await Promise.all([
    fs.readFile("./src/i18n/translations/legacy/en.yml", "utf8"),
    fs.readFile("./src/i18n/translations/legacy/ja.yml", "utf8"),
    fs.readFile("./src/i18n/translations/ja.yml", "utf8"),
  ]);
  const legacyEn = yaml.load(legacyEnRaw);
  const legacyJa = yaml.load(legacyJaRaw);
  const newJa = yaml.load(newJaRaw);

  const result = {};
  const keys = Object.keys(newJa);
  for (const key of keys) {
    // If value is only exist in new ja file then use existing value.
    if (newJa[key]) {
      result[key] = newJa[key];
      continue;
    }

    const replacedKey = key.replace(/\{\{(.+)\}\}/, "{$1}");
    const hash = findKeyByValue(legacyEn, replacedKey);
    if (hash) {
      const value = findValueByKey(legacyJa, hash);
      result[key] = value;
    } else {
      throw new Error("Hash is not found");
    }
  }

  const rawYaml = yaml.dump(result);
  await fs.writeFile("./src/i18n/translations/ja.yml", rawYaml, "utf8");
};

run();
</code></pre>
</details>


keiya01:
[fix: use key as english translation](https://github.com/reearth/reearth-web/pull/240/commits/b83c3a55e6ebcff8a33244cd58c4cc227f4ddc11)
I think we can use translation object key as translation text in english.
For example, when we use `useT("Hello World")` then translation key is `Hello World`. And I think in most case, this key is actual text used in web page.
Therefore we don't need to define `en.yml` translation file, we can use key as translation text in english.
When user specify `ja` lang, service text is displayed in Japanese.
Do @rot1024 @KaWaite  think about this?

package.json
```diff
@@ -147,7 +144,8 @@
         "webpack": "^5.72.1",
         "webpack-cli": "^4.9.2",
         "webpack-dev-server": "^4.9.0",
-        "yaml-flat-loader": "^1.0.1"
+        "yaml-flat-loader": "^1.0.1",
```
rot1024, line 147:
```suggestion
```

replace_i18n.mjs
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      throw new Error("Hash is not found");
+    }
+  }
+
+  const rawYaml = yaml.dump(result);
+  await fs.writeFile("./src/i18n/translations/ja.yml", rawYaml, "utf8");
+};
+
+run();
```
rot1024, line 85: Thank you for implementing a nice script! but you can delete it since it is no longer used in the future.

i18next-parser.config.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
-  locales: ["en", "ja"],
+  locales: ["ja"],
```
rot1024, line 2:
```suggestion
  locales: ["en"],
```

I would prefer to generate `en.yml` as well.

This problem is also a question of how to name the keys. We have adopted the method of using the English text as the key. The advantage of this is that we do not have to write the English text again in the YAML file. The disadvantage is not so much that it is difficult to translate the same English text into different texts.

By generating `en.yml` and leaving the keys lined up but leaving the translations empty, you can write the English text in the source code and when the same English text is used, you can change the key names slightly and write the correct English text in the `en.yml` file. If the key is empty, the key will be displayed on the screen as it is, according to the i18next settings.

rot1024 (changes requested):
- Delete `i18n:legacy` script in `package.json` also. 
- Resolve conflict, but It would not be a problem to override change in this PR.
- CI failed. CI is set up to report an error if there are missing translation. Rerun `yarn i18n`

replace_i18n.mjs
keiya01: oh... I didn't intend to commit these script. I removed these script at this commit.
[Revert "fix: use key as english translation"](https://github.com/reearth/reearth-web/pull/240/commits/ff56843132be0b8d85e5eb974d479901e9940ebf)

i18next-parser.config.js
keiya01: I reverted `en.yml` at this commit :pray:
[Revert "fix: use key as english translation"](https://github.com/reearth/reearth-web/pull/240/commits/ff56843132be0b8d85e5eb974d479901e9940ebf)

package.json
keiya01: I removed unnecessary script from package.json.
[refactor: remove yarn-flat-loader](https://github.com/reearth/reearth-web/pull/240/commits/923c78453d166482d5439369804b452d86905f5f)
[refactor: remove unnecessary script](https://github.com/reearth/reearth-web/pull/240/commits/6351eed1bc1425350c0b6c91b2d092cb1b86bf00)

KaWaite: One of your commits after the `remove yarn-flat-loader` commit brought it back. Please delete it again

package.json
```diff
@@ -147,7 +143,7 @@
         "webpack": "^5.72.1",
         "webpack-cli": "^4.9.2",
         "webpack-dev-server": "^4.9.0",
-        "yaml-flat-loader": "^1.0.1"
+        "yaml-loader": "^0.8.0"
```
KaWaite, line 146: This is suppose to be removed, right?

src/i18n/i18n.ts
```diff
@@ -20,7 +20,11 @@ export const availableLanguages = Object.keys(resources);
 i18n.use(LanguageDetector).use(initReactI18next).init({
   resources,
   fallbackLng: "en",
+
+  // allow keys to be phrases having `:`, `.`
+  nsSeparator: false,
   keySeparator: false,
+
```
KaWaite, line 27: Not sure there is a reason you spaced things out a bit, but since they are an object being passed in to `init`, please keep things tight.
```suggestion
  // allow keys to be phrases having `:`, `.`
  nsSeparator: false,
  keySeparator: false,
```

src/i18n/i18n.ts
```diff
@@ -20,7 +20,11 @@ export const availableLanguages = Object.keys(resources);
 i18n.use(LanguageDetector).use(initReactI18next).init({
   resources,
   fallbackLng: "en",
+
```
KaWaite, line 23: Since we are passing an object in to `init` here, please remove the empty spaces and keep things tight

webpack.config.js
```diff
@@ -102,8 +102,7 @@ module.exports = (env, args = {}) => {
         },
         {
           test: /\.ya?ml$/,
-          // Delete yaml-flat-loader and use a regular yaml loader when migration to i18next is complete
-          use: [{ loader: "json-loader" }, { loader: "yaml-flat-loader" }],
+          loader: "yaml-loader",
```
KaWaite, line 105: Is it okay you deleted `json-loader` here along with `yaml-flat-loader`? The comment said to only remove the yaml flat loader and replace it with yaml-loader.

KaWaite (approved):
Looked over the config files only, but everything looks good! I asked for a single change, so please do that before merging. 

Also, I'll approve, but can you please test storybook locally? The PR's preview deploy on netlify failed. Sometimes changes to yarn.lock, especially deleting yarn.lock and regenerating, will break storybook.... 
If it fails on local as well, please investigate, fix and ask for a re-review! If it works locally it might be an issue with redeployment on netfliy and should be okay for us.

src/i18n/i18n.ts
keiya01: https://github.com/reearth/reearth-web/pull/240/commits/7422aeaf7aa11eae174bacd235f1d5420f434829
I removed space 🙏

rot1024 (approved):

PR #245, opened 2022-06-10
feat: change layer indicators from preset list

nina992:
# Overview
Layer selecting indicator: is an indicator appears when user select a layer.
We  created preset list of indicators that user can choose one of them when selecting a layer.

## What I've done
Design a preset list of indicators for layer selecting.
Add indicator section in scene right tab and give the user the ability to choose among them.
Give the user the ability to upload his own indicator.

## Screenshot
image
image
image
image

## Memo
Design link : https://www.figma.com/file/bdnfDaGXGoNfUws4y7JItl/Re%3AEarth-UI-Gamma?node-id=10319%3A174052
Backend PR: https://github.com/reearth/reearth-backend/pull/158


src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -88,6 +93,11 @@ export type SceneProperty = {
     cameraLimitterTargetWidth?: number;
     cameraLimitterTargetLength?: number;
   };
+  indicator?: {
+    indicator_type: indicatorTypes;
+    indicator_image?: string;
+    img_scale?: number;
```
rot1024, line 99:
```suggestion
    indicator_image_scale?: number;
```

Rename it because it should be unique among one property schema. `img_scale` is too abstract and it is unclear what the image is. Also manifest on reearth-backend should be updated.

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -22,6 +22,11 @@ export type EngineRef = {
 };
 
 export type SceneMode = "3d" | "2d" | "columbus";
+export enum indicatorTypes {
+  default = "default",
+  crosshair = "crosshair",
+  custom = "custom",
+}
```
rot1024, line 29:
```suggestion
export type indicatorTypes = "default" | "crosshair" | "custom";
```

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: rgba(0, 0, 0, 0.5);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  transition: ${({ transition }) =>
+    transition === "entering" || transition === "exiting" ? "all 0.5s ease" : ""};
+  opacity: ${({ transition }) => (transition === "entering" || transition === "entered" ? 1 : 0)};
```
rot1024, line 132: Is it correct that the animation should only fade in and out? What about animations that change scale? @lavalse

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/ref.ts
nina992: done

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+      }
+    };
+
+    viewer.clock.onTick.addEventListener(handleTick);
+    return () => {
+      if (viewer.isDestroyed()) return;
+      viewer.clock.onTick.removeEventListener(handleTick);
+    };
+  }, [viewer]);
+  // console.log( indicator_type );
```
KaWaite, line 74: Please remove

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return transiton !== "unmounted" && pos ? (
+    indicator_type === "crosshair" ? (
+      <StyledIcon
+        icon="crosshair"
+        className={className}
+        transition={transiton}
+        style={{ left: pos.x + "px", top: pos.y + "px" }}
+      />
+    ) : indicator_type === "default" ? (
```
KaWaite, line 83: Shouldn't the cesium default be the fallback option, not the Image?

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+        width={w}
+        height={h}
+        transition={transiton}
+        style={{ left: pos.x + "px", top: pos.y + "px" }}
+      />
+    )
+  ) : null;
+}
+
+const I = styled.div<{ transition: TransitionStatus }>`
```
KaWaite, line 101: `I` has no obvious meaning. Please rename it so it is easier to understand what this component is. If it is an Indicator, please just call it an `Indicator`.

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+  height: 50px;
+  transform: translate(-50%, -50%);
+  transition: ${({ transition }) =>
+    transition === "entering" || transition === "exiting" ? "all 0.5s ease" : ""};
+  opacity: ${({ transition }) => (transition === "entering" || transition === "entered" ? 1 : 0)};
+`;
+const StyledIcon = styled(Icon)<{ transition: TransitionStatus }>`
+  position: absolute;
+  width: 50px;
+  height: 50px;
```
KaWaite, line 113: Does the image scale correctly with this? Usually with the Icon component you need to pass it a `size` prop and the component handles the scaling. This is especially important with SVGs. Please make sure the StyledIcon is actually scaling correctly (in this case 50X50) and not just assuming it is, because it might be smaller/bigger than it should be.

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+  transform: translate(-50%, -50%);
+  transition: ${({ transition }) =>
+    transition === "entering" || transition === "exiting" ? "all 0.5s ease" : ""};
+  opacity: ${({ transition }) => (transition === "entering" || transition === "entered" ? 1 : 0)};
+`;
+const Image = styled.img<{ transition: TransitionStatus }>`
```
KaWaite, line 119:
```suggestion
`;

const Image = styled.img<{ transition: TransitionStatus }>`
```

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+    indicator_type === "crosshair" ? (
+      <StyledIcon
+        icon="crosshair"
+        className={className}
+        transition={transiton}
+        style={{ left: pos.x + "px", top: pos.y + "px" }}
+      />
+    ) : indicator_type === "default" ? (
+      <I
+        className={className}
```
KaWaite, line 85: Is this className necessary? Also, from the code I can't see this `I` component displaying anything. So it seems like just an empty div. Am I wrong?

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -60,6 +60,11 @@ export default ({
   // imagery layers
   const [imageryLayers, setImageryLayers] = useState<ImageryLayerData[]>();
 
+  //indicator
```
KaWaite, line 63:
```suggestion
  // indicator
```

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -22,6 +22,7 @@ export type EngineRef = {
 };
 
 export type SceneMode = "3d" | "2d" | "columbus";
+export type indicatorTypes = "default" | "crosshair" | "custom";
```
KaWaite, line 25: export type IndicatorTypes = "default" | "crosshair" | "custom";

Please keep consistent with spacing, naming, etc. As you can see above all types are capitalized.

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+import Icon from "@reearth/components/atoms/Icon";
+import { styled } from "@reearth/theme";
+
+import { SceneProperty } from "../../ref";
+import { useIcon } from "../common";
+
+export type Props = {
+  className?: string;
+  property: SceneProperty | undefined;
```
KaWaite, line 14:
```suggestion
  property?: SceneProperty;
```

KaWaite (commented):
A few comments.
Also, is there any chance we will support changing the color of the crosshair indicator? @lavalse maybe can shed some light on that. Currently the svg is setup to only be white, but would be pretty simple to allow for color.

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
rot1024: Here is the code I wrote. Basically, I'm trying to name the class so that you can edit the style with a styled-components type library. You can delete it.

src/components/molecules/Visualizer/Engine/ref.ts
rot1024: Sorry it's my suggestion

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
nina992: @KaWaite  please check is that what you mean ?

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
nina992: Done

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
KaWaite: But here the logic goes:
- is it crosshair? show crosshair. if not,
- is it default? show default. if not,
- show Image component

So the fallback is the Image component (`indicator_type === "custom"` right?). Can you make it so that `indicator_type === "custom"` and Image component is checked, and then it fallsback to the default?

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -60,6 +60,16 @@ export default ({
...
 
+  // indicator
+  const selectionIndicator = useMemo(() => {
+    if (
+      property?.indicator?.indicator_type == undefined ||
+      property?.indicator?.indicator_type == "default"
+    )
+      return true;
+    return false;
+  }, [property?.indicator?.indicator_type]);
```
KaWaite, line 71: This can be cleaned up like so:
```
  const selectionIndicator = useMemo(() => {
!!(property?.indicator?.indicator_type == undefined || property?.indicator?.indicator_type == "default");
  }, [property?.indicator?.indicator_type]);
```

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
nina992: @lavalse  would please check

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
nina992: @KaWaite  agreed on this right?

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -60,6 +60,24 @@ export default ({
...
 
+  // indicator
+  // const selectionIndicator = useMemo(() => {
+  //   if (
+  //     property?.indicator?.indicator_type == undefined ||
+  //     property?.indicator?.indicator_type == "default"
+  //   )
+  //     return true;
+  //   return false;
+  // }, [property?.indicator?.indicator_type]);
```
KaWaite, line 71: Don't need this anymore :P

lavalse:
Sorry for testing a little bit late. But I found some problems maybe more importent then animation. 😄
- if use indicater which is not default one, and if the curser on the indicator, it can not be zoom
- the marker lost Drag and drop feature
Not sure how to solve, I feel maye a little difficult

KaWaite (approved):
LGTM

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
nina992: @rot1024  , please note that red add his comment and we fixed pr accordingly so I will resolve this one , is that ok ?

src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
nina992: @KaWaite  can I resolve this one

rot1024 (approved):

PR #249, opened 2022-06-15
fix: emit update event of plugin API by scene property update

rot1024:
Emit update  event of plugin API by scene property update


rot1024:
A problem was found in which the sceneProperty referenced by the plugin API remained out of date even if the plugin was successfully notified of an update of the sceneProperty. Fixing this will take some time, so it will be done at the same time as the layer system will be revamped in the future.

PR #250, opened 2022-06-20
fix: router typo

airslice:
# Overview

some route target path are missing updates after the router got updated.

## What I've done

fix path `workspace` -> `workspaces`
fix path `project` -> `projects`

## How I tested

Visting these page:

setting page for project 
- Public
- Dataset
- Plugins

settings page:
- ProjectList
- Assets

One note from the router updates:
Some urls are changed in order to make them more semantic. If someone bookmark and visit the former urls for these pages will get a 404.


rot1024 (commented):
Can you add redirections? e.g. `/settings/workspace/*` -> `/settings/workspaces/*`

PR #251, opened 2022-06-20
test: router updates

airslice:
# Overview

This is only a draft for test.

I notice the preview version keeps blank and come up with errors when i do the fixing for router in another PR. 

This draft PR did nothing but add a comment so it should be same with the main branch. 

But the deploy log shows warnings and the preview page has error too.

Please check the preview deploy log.


airslice:
Solved by clear cache of netlify

PR #252, opened 2022-06-20
fix: router typos

airslice:
# Overview

Some route target path of the Navigation component didn't got update after the router path got updated. 

These pages are 404 from navigation:

Workspace setting pages:
- Workspace
- Assets
- ProjectList

Project setting pages: 
- Project
- Public
- Dataset
- Plugins

## What I've done

- fix path `workspace` -> `workspaces`
- fix path `project` -> `projects`
- add redirection for the outdated urls.

- also moved `why-did-you-render` by @rot1024 

## How I tested

Visiting the pages list above from UI navigation.
Visiting the outdated urls directly.

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -1,5 +1,11 @@
 import React, { Suspense } from "react";
-import { BrowserRouter as Router, useRoutes, Navigate, useParams, Params } from "react-router-dom";
+import {
+  BrowserRouter as Router,
+  useRoutes,
+  Navigate,
+  useParams,
+  type Params,
+} from "react-router-dom";
 
 import Loading from "@reearth/components/atoms/Loading";
 import NotificationBanner from "@reearth/components/organisms/Notification";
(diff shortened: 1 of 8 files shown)


src/app.tsx
```diff
@@ -60,6 +68,35 @@ function AppRoutes() {
     { path: "/settings/projects/:projectId/plugins", element: <PluginSettings /> },
     { path: "/plugin-editor", element: <PluginEditor /> },
     { path: "/graphql", element: process.env.NODE_ENV !== "production" && <GraphQLPlayground /> },
+    // redirections for breaking changes in urls
+    {
+      path: "/settings/workspace/:teamId",
+      element: <Redirect to="/settings/workspaces/:teamId" />,
```
KaWaite, line 74: Are these really necessary? I feel like we are making things more complex here to avoid having to update URLs in probably just the documentation website.
I think remember seeing @rot1024 asked for these, so maybe he can shed some light here?

airslice: As we talked i want to provide the redirections at least for the next release so that users can be noticed about the breaking changes from release note -> update outdated urls in bookmark or docs (if any) and won't facing a 404 directly. The codes from @rot1024 also shows he want to keep it maybe. So i just mark the coversation as resolved now and let me know if you want to discuss more :)

PR #254, opened 2022-06-20
fix: dependency @emotion/utils error

airslice:
# Overview

When deploy preview on netlify it come up with warning:

`WARNING in ./node_modules/react-spinners/node_modules/@emotion/react/dist/emotion-element-cbed451f.browser.esm.js 183:2-16
export 'registerStyles' (imported as 'registerStyles') was not found in '@emotion/utils' (possible exports: getRegisteredStyles, insertStyles)`

And the preview page throw error:

`vendor-c0b5c364.108108bf.js:2 TypeError: (0 , r.registerStyles) is not a function`

This makes the preview not working.

More discussions can be found [here](https://github.com/emotion-js/emotion/issues/2702).

## What I've done

Add "resolutions" for @emotion/utils to force using v1.1.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,7 @@
         "use-file-input": "^1.0.0"
     },
     "resolutions": {
-        "@emotion/utils": "^1.1.0"
+        "@emotion/utils": "^1.1.0",
+        "@emotion/sheet": "^1.1.0"
     }
 }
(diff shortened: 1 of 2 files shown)


airslice:
All these actually don't work. 
Actually it can be fixed by clean netlify depoloy cache and the new branches created after the big upgrade PR seems work normally. 
This fix is no longer needed.

PR #255, opened 2022-06-21
fix: auto fetch more items in dashboard page , project list , dataset page for big screens

nina992:
# Overview
After adding infinite scroll and for big monitors the page only shows limit number of items and the scroll bar didn't appear so the system didn't fetch all items related to this list.

## What I've done
fix this and auto fetch the items when the page is too large


src/util/AutoFetchMore.ts
```diff
@@ -0,0 +1,10 @@
+import { MutableRefObject } from "react";
+
+export function AutoFetchMore(
```
issmail-basel, line 3: Is there a reason why this function starts with a capital letter?
I think it should start with small

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -97,7 +98,10 @@ const AssetContainer: React.FC<Props> = ({
     onRemove,
     onSearch,
   });
-
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    if (wrapperRef.current && !isLoading && hasMoreAssets) AutoFetchMore(wrapperRef, onGetMore);
+  }, [hasMoreAssets, isLoading, onGetMore]);
```
KaWaite, line 104: Please think about how things are spaced out. We don't have a definitive rule because the types of files/components are various, but trying to space things well is important for readability. In this case, useT, useHooks and useRef are all declaring variables being used elsewhere, so is typically fine side by side, but useEffects are unique and I think need to stand on their own so it is easier to spot and acknowledge.
Hope that makes sense. Please ask me on slack if you want more of an explanation or you want to debate it!
```suggestion
  const wrapperRef = useRef<HTMLDivElement>(null);
  
  useEffect(() => {
    if (wrapperRef.current && !isLoading && hasMoreAssets) AutoFetchMore(wrapperRef, onGetMore);
  }, [hasMoreAssets, isLoading, onGetMore]);
  
```

src/components/molecules/Settings/SettingPage/index.tsx
```diff
@@ -29,7 +30,10 @@ const SettingPage: React.FC<SettingPageProps> = ({
   const handleClick = () => {
     setIsOpen(o => !o);
   };
-
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    if (wrapperRef.current && !loading && hasMoreItems) AutoFetchMore(wrapperRef, onScroll);
+  }, [hasMoreItems, loading, onScroll]);
```
KaWaite, line 36:
```suggestion

  const wrapperRef = useRef<HTMLDivElement>(null);
  
  useEffect(() => {
    if (wrapperRef.current && !loading && hasMoreItems) AutoFetchMore(wrapperRef, onScroll);
  }, [hasMoreItems, loading, onScroll]);
  
```

src/util/AutoFetchMore.ts
```diff
@@ -0,0 +1,10 @@
...
+  ref: MutableRefObject<HTMLDivElement | undefined | null>,
+  onLoadMore?: () => void,
+) {
+  if (ref.current && ref.current?.scrollHeight <= document.documentElement.clientHeight) {
+    onLoadMore?.();
+  }
+}
```
KaWaite, line 10: Since they are related, can you move this and `handleScroll` into the same file? Since they both relate to scrolling, the file could be called scroll.ts.
Then, please rename them to be more appropriate revolving around scrolling. 
ex.
handleScroll -> onScrollToBottom
AutoMergeFetch -> autoFillPage

These names aren't great so if you can think of better names for them, that are general, please do. Or, one thing I can't decide is whether these components are too specific to infinite scroll. If so, we could move them both into an infinite-scroll.ts file in the Common directory. What do you think?

src/components/molecules/Common/AssetModal/AssetContainer/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useRef } from "react";
```
KaWaite, line 1:
```suggestion
import { useEffect, useRef } from "react";
```

src/util/AutoFetchMore.ts
nina992: I will change the name totally

src/util/AutoFetchMore.ts
nina992: I think infinte-scroll.ts is suitable as I don't think we can use it any where else

src/util/infinite-scroll.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function autoFillPage(
+  ref: MutableRefObject<HTMLDivElement | undefined | null>,
+  onLoadMore?: () => void,
+) {
+  if (ref.current && ref.current?.scrollHeight <= document.documentElement.clientHeight) {
+    onLoadMore?.();
+  }
+}
+export function onScrollToBottom(
```
KaWaite, line 11:
```suggestion
}

export function onScrollToBottom(
```

src/components/molecules/Dashboard/index.tsx
```diff
@@ -23,11 +23,17 @@ const Dashboard: React.FC<Props> = ({
   hasMoreProjects,
   onGetMoreProjects,
 }) => {
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    if (wrapperRef.current && !isLoading && hasMoreProjects)
+      autoFillPage(wrapperRef, onGetMoreProjects);
+  }, [hasMoreProjects, isLoading, onGetMoreProjects]);
```
KaWaite, line 30:
```suggestion
  const wrapperRef = useRef<HTMLDivElement>(null);
  
  useEffect(() => {
    if (wrapperRef.current && !isLoading && hasMoreProjects)
      autoFillPage(wrapperRef, onGetMoreProjects);
  }, [hasMoreProjects, isLoading, onGetMoreProjects]);
  
```

KaWaite (approved):

PR #258, opened 2022-06-24
feat: plugin API to add layers

airslice:
close https://github.com/reearth/reearth/issues/306

# Overview

This new feature will enable plugin add new layer to reearth through API.

Add new layer can be done by call `reearth.layers.add(layer: Layer, parentId?: string)`

- The layer created by API is actually an instance of the built-in plugin `reearth` with different extensionIds, therefore its `pluginId` is always `reearth` and there's no need to include `pluginId` in the layer object.
- The layer object has defferent schema with different `extensionIds`. Currently the schema can only ref from backend and maybe will doc it somewhere later.
- It's easier to use external url for static resouces (img, model, etc.). Use reearth assets is also ok but you need to upload them from reearth editor and get the asset link first.
- The infobox of layer is also supported but as there's no override method to it i think it can be set only once.

Maybe this implementation will not last for a long time as the layer system is under redesigning now.

## What I've done

Extension for plugin API:

 - `reearth.layers.add(layer:Layer, parentId?:string) => string` 

Structure changes:

- The `layerStore` now is managed in molecules Visualizer (no longer in organisms level);
- The `layerStore` now is reused (only created when initial);
  - The layers data now has two parts:
    - `databaseLayers` -> the root layer from database. When got data from gql the `setDatabaseLayers` will be called and all layers will be updated. 
    - `appendedLayersData` -> the array of layers data from API. 
  - The database layers will be treated as first class member in order to support behavior like append layers to folder from database.
  - The layer created by API will get a unique id which has the same length with the database layer but start with "_".
  - Layer data in tree structure is also supported.
  - The `Layer` has a new property `creator` to indicate if it's created by API.

## What I haven't done

- No `remove` method in this PR.

## How I tested

Made a [test plugin](https://github.com/airslice/reearth-plugin-api-demo/blob/main/packages/reearth-plugin-api-demo-0.0.1.zip) and use it to:

- [x] get extensionIds
- [x] add single marker
- [x] add single photooverlay
- [x] add single ellipsoid
- [x] add single model
- [x] add single tileset
- [x] add single resouce
- [x] add single polyline
- [x] add single polygon
- [x] add single rect
- [x] override property to appended layer
- [x] add folder
- [x] add layer(use marker as example) to folder
- [x] add layer(use marker as example) with infobox
- [x] add layer(use marker as example) with tree structure

## Screenshot
image

- For the component update:
  - I still use the additional state `layersRenderKey` to trigger an update since it work fine and `useUpdate` cannot be used inside useCallback;

- Inside the layerStore class, in order get the latest layers data the flattern and map functions got called every time append layer or setDatabaseLayers. I'm not sure if can simplify it.


src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,70 @@ export class LayerStore {
...
+      },
+      parentId,
+    };
+    if (layer.infobox?.blocks) {
+      layer.infobox.blocks.map(b => {
+        b.id = this.#getUniqueAppendedResId();
+      });
+    }
+    this.#appendedLayersData.push(appendedLayerData);
+    this.#insertAppendedLayer(appendedLayerData);
```
rot1024, line 131: Not including `this.#appendedLayersData.push(appendedLayerData);` in the `#insertAppendedLayer ` function seems a little hard to read in the code, I think

rot1024 (changes requested):

src/components/molecules/Visualizer/Plugin/context.tsx
```diff
@@ -108,6 +110,15 @@ export function Provider({
   const getLayerSelectionReason = useGet(layerSelectionReason);
   const getLayerOverriddenInfobox = useGet(layerOverridenInfobox);
   const getLayerOverriddenProperties = useGet(layerOverriddenProperties);
+  const getBuiltinExtensionIds = useCallback(
+    () =>
+      builtinPrimitives
+        ? Object.keys(builtinPrimitives)
+            .filter(p => !["reearth/polyline", "reearth/polygon", "reearth/rect"].includes(p))
```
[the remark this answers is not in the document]
airslice: okey i will test these types.

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -210,31 +239,47 @@ export default ({
...
-  layers: LayerStore;
+  rootLayer: Layer | undefined;
   selected?: string;
   onSelect?: (id?: string, options?: SelectLayerOptions) => void;
 }) {
   const [selectedLayerId, innerSelectLayer] = useState<string | undefined>();
   const [layerSelectionReason, setSelectionReason] = useState<string | undefined>();
   const [layerOverridenInfobox, setPrimitiveOverridenInfobox] = useState<OverriddenInfobox>();
+  const [layers] = useState<LayerStore>(() => new LayerStore(rootLayer));
+  const [, setLayersRenderKey] = useState<number>(0);
```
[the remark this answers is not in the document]
airslice: Actually just because i did not know how to use it inside a useCallback. I've updated this part please help have a look if i use it correctly.

src/components/organisms/EarthEditor/CanvasArea/convert.ts
```diff
@@ -145,7 +145,7 @@ type GQLLayer = Maybe<
     EarthLayerCommonFragment
 >;
 
-const processLayer = (layer: Maybe<GQLLayer>): Layer | undefined => {
+export const processLayer = (layer: Maybe<GQLLayer>): Layer | undefined => {
```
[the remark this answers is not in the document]
airslice: I left it because i was not sure if the whole plan (move layerStore creation to molecules) is okey and if not i still can change it back quickly. I'm deleting it now.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,70 @@ export class LayerStore {
...
+
+  #updateLayers = () => {
+    this.#root = this.#databaseLayers ?? { id: "", children: [] };
+    this.#proot = this.#pluginLayer(this.#root);
+    this.#mapLayers();
+    this.#appendedLayersData.map(l => this.#insertAppendedLayer(l));
+  };
+
+  #appendedResIds: string[] = [];
+  #getUniqueAppendedResId = () => {
```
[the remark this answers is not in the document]
airslice: Updated.

src/components/organisms/EarthEditor/CanvasArea/hooks.ts
```diff
@@ -99,8 +99,20 @@ export default (isBuilt?: boolean) => {
...
-  const selectedLayer = selectedLayerId ? layers?.findById(selectedLayerId) : undefined;
+
+  const rootLayer = useMemo(
+    () =>
+      !(
+        !layerData ||
+        !layerData.node ||
+        layerData.node.__typename !== "Scene" ||
+        !layerData.node.rootLayer
+      )
```
KaWaite, line 110: Maybe this is just personal preference, but I feel like its harder to understand whats happening doing `!(!something || !somethingelse .. )` than doing `!!(something || somethingelse...)`. That way, there is less negating and converting we have to do in our heads.
If you agree, could this be re-written like: 
```suggestion
      !!(
        layerData?.node ||
        layerData.node.__typename === "Scene" ||
        layerData.node.rootLayer
      )
```
I removed the first `layerData ||` and just add the question mark to the first check. Please make sure that is actually okay.

src/components/molecules/Visualizer/Plugin/types.ts
```diff
@@ -74,6 +74,8 @@ export type Layers = {
   ) => T | undefined;
   readonly isLayer: (obj: any) => obj is Layer;
   readonly overrideProperty: (id: string, property: any) => void;
+  readonly append: (layer: Layer, parentId?: string) => string;
+  readonly extensionIds: string[];
```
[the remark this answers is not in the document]
airslice: - Currently the `append` method accept only one layer and its parentId as option so it just do append to root layer or parent layer. But i think it should support tree structure as it was designed. I'm going to update this part and if it works update the method name to `add`.
- The `getExtensionIds` API comes from the former talk. I agree that this can be done in document and i'm going to remove it.

src/components/molecules/Visualizer/Plugin/context.tsx
```diff
@@ -108,6 +110,14 @@ export function Provider({
   const getLayerSelectionReason = useGet(layerSelectionReason);
   const getLayerOverriddenInfobox = useGet(layerOverridenInfobox);
   const getLayerOverriddenProperties = useGet(layerOverriddenProperties);
+  const getBuiltinExtensionIds = useCallback(
+    () =>
+      builtinPrimitives
```
KaWaite, line 115: Getting a CI linting warning here. Make sure to save locally this file if you accepted changes by @rot1024 directly in github. Sometimes the formatting isn't correct.

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -210,31 +238,47 @@ export default ({
 };
 
 function useLayers({
-  layers,
+  rootLayer,
   selected: outerSelectedPrimitiveId,
   onSelect,
 }: {
-  layers: LayerStore;
+  rootLayer: Layer | undefined;
```
KaWaite, line 245: Just to keep the writing style of props in this file consistent:
```suggestion
  rootLayer?: Layer;
```

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -210,31 +238,47 @@ export default ({
...
-    () => (selectedLayerId ? layers.findById(selectedLayerId) : undefined),
+    () => (selectedLayerId ? layers?.findById(selectedLayerId) : undefined),
     [selectedLayerId, layers],
   );
 
   const selectLayer = useCallback(
     (id?: string, { reason, overriddenInfobox }: SelectLayerOptions = {}) => {
       innerSelectLayer(id);
-      onSelect?.(id);
+      id && !!layers.findById(id) && onSelect?.(id);
```
KaWaite, line 268: Is it okay, for example, if `!!layers.findById(id)` is false for the other methods inside this useCallback to be called without `onSelect`?
Or is it better to move this check to the top of the useCallback like:
`if (!id || !layers.findById(id)) return;` so that its all or nothing?

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -243,20 +242,20 @@ function useLayers({
...
   onSelect?: (id?: string, options?: SelectLayerOptions) => void;
 }) {
   const [selectedLayerId, innerSelectLayer] = useState<string | undefined>();
   const [layerSelectionReason, setSelectionReason] = useState<string | undefined>();
   const [layerOverridenInfobox, setPrimitiveOverridenInfobox] = useState<OverriddenInfobox>();
   const [layers] = useState<LayerStore>(() => new LayerStore(rootLayer));
-  const [, setLayersRenderKey] = useState<number>(0);
+  const forceUpdate = useUpdate();
 
   useMemo(() => {
```
rot1024, line 255:
```suggestion
  useEffect(() => {
```
Incorrect usage of useMemo

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+        return ((Math.random() * 16) | 0).toString(16);
+      });
+    let id;
+    do {
+      id = genResId();
+    } while (this.#usedIds.includes(id));
+    this.#usedIds.push(id);
+    return id;
+  };
+
```
rot1024, line 89:
```suggestion
  #usedIds: Set<string> = new Set();
  #newId = () => {
    const genResId = () =>
      "_xxxxxxxxxxxxxxxxxxxxxxxxx".replace(/[x]/g, function () {
        return ((Math.random() * 16) | 0).toString(16);
      });
    let id;
    do {
      id = genResId();
    } while (this.#usedIds.has(id));
    this.#usedIds.add(id);
    return id;
  };

```

In this use case, Set is extremely faster than Array. Element retrieval in an array is O(n) while element retrieval in a Set is O(1). As the number of elements increases, the performance difference is obvious.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+    const { children, ...self } = layer;
+    self.infobox?.blocks &&
+      self.infobox.blocks.forEach(b => {
+        b.id = this.#newId();
+      });
+    return {
+      ...self,
+      id: this.#newId(),
+      pluginId: "reearth",
+      creator: "pluginAPI",
```
rot1024, line 113: Instead of a fixed string, creator should be the ID of the plugin. Then we can implement a method to bulk delete layers added by a particular plugin. This is useful for cleanup when a plugin is destroyed.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+  };
+
+  #processLayer = (layer: Omit<Layer, "id">): Layer => {
+    const { children, ...self } = layer;
+    self.infobox?.blocks &&
+      self.infobox.blocks.forEach(b => {
+        b.id = this.#newId();
+      });
+    return {
+      ...self,
```
rot1024, line 110:
```suggestion
    return {
      ...self,
      ...(self.infobox ? {
        infobox: {
          ...self.infobox,
          blocks: self.inbox.blocks?.map(b => ({ ...b, id: this.#newId() }))
        }
      } : {}),
```

It is safer to recreate the object unless performance is a concern, as changing the original object can lead to unexpected bugs. Also, the implementation is not complete if the value being returned is an object that has been rebuilt, but only the blocks have not been rebuilt.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+    const id = this.#insertLayer(
+      {
+        layer: this.#processLayer(layer),
+        parentId,
+      },
+      true,
+    );
+    this.#proot = this.#pluginLayer(this.#root);
+    this.#flattenLayers = flattenLayers(this.#root?.children ?? []);
+    return id;
```
rot1024, line 128:
```suggestion
  add = (layer: Omit<Layer, "id">, parentId?: string, creator?: string) => {
    const id = this.#insertLayer(
      {
        layer: this.#processLayer(layer, creator),
        parentId,
      },
      true,
    );
    this.#proot = this.#pluginLayer(this.#root);
    this.#flattenLayers = flattenLayers(this.#root?.children ?? []);
    return id;
```

You may want to receive `creator` here.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+
+  add = (layer: Omit<Layer, "id">, parentId?: string) => {
+    const id = this.#insertLayer(
+      {
+        layer: this.#processLayer(layer),
+        parentId,
+      },
+      true,
+    );
+    this.#proot = this.#pluginLayer(this.#root);
```
rot1024, line 126:
```suggestion
```

If you remove this code, it should still work.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+  #insertLayer = (layerData: AddedLayerData, save = false) => {
+    const tar = layerData.parentId ? this.#map.get(layerData.parentId) : this.#root;
+    if (!tar?.children) return;
+    tar.children.push(layerData.layer);
+    const flattenedLayers = flattenLayers([layerData.layer]);
+    flattenedLayers.forEach(l => {
+      this.#map.set(l.id, l);
+      this.#pmap.set(l.id, this.#pluginLayer(l));
+    });
+    save && !tar.creator && this.#addedLayersData.push(layerData);
```
rot1024, line 99:
```suggestion
    if (save && !tar.creator) this.#addedLayersData.push(layerData);
```

- IMO (I like if statement in normal codes, except JSX)
- Tell me why `!tar.creator` is needed

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+
+  #insertLayer = (layerData: AddedLayerData, save = false) => {
+    const tar = layerData.parentId ? this.#map.get(layerData.parentId) : this.#root;
+    if (!tar?.children) return;
+    tar.children.push(layerData.layer);
+    const flattenedLayers = flattenLayers([layerData.layer]);
+    flattenedLayers.forEach(l => {
+      this.#map.set(l.id, l);
+      this.#pmap.set(l.id, this.#pluginLayer(l));
+    });
```
rot1024, line 98:
```suggestion
    flattenLayers([layerData.layer]).forEach(l => {
      this.#map.set(l.id, l);
      this.#pmap.set(l.id, this.#pluginLayer(l));
    });
```

If a variable is defined, expect it to be reused later.

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -68,6 +73,72 @@ export class LayerStore {
...
+        parentId,
+      },
+      true,
+    );
+    this.#proot = this.#pluginLayer(this.#root);
+    this.#flattenLayers = flattenLayers(this.#root?.children ?? []);
+    return id;
+  };
+
+  setDatabaseLayers = (databaseRootLayer: Layer | undefined) => {
```
rot1024, line 131:
```suggestion
  setRootLayer = (root: Layer | undefined) => {
```

IMO: simpler name

src/components/molecules/Visualizer/Layers/store.ts
```diff
@@ -47,9 +49,11 @@ export class LayerStore {
...
-    this.#root = root;
-    this.#proot = this.#pluginLayer(root);
-    this.#flattenLayers = flattenLayers(root?.children ?? []);
+    this.#root = databaseRootLayer ?? { id: "", children: [] };
+    this.#addedLayersData = [];
+
+    this.#proot = this.#pluginLayer(this.#root);
+    this.#flattenLayers = flattenLayers(this.#root?.children ?? []);
     this.#map = new Map(this.#flattenLayers.map(l => [l.id, l]));
     this.#pmap = new Map(this.#flattenLayers.map(l => [l.id, this.#pluginLayer(l)]));
```
rot1024, line 58:
```suggestion
    this.setRoot(databaseRootLayer);
```

Can we put together a common process like this?

rot1024 (changes requested):

rot1024 (approved):

KaWaite (approved):

PR #263, opened 2022-07-04
fix: testing-library react 18 warnings

KaWaite:
# Overview
When executing `yarn test` we were getting this error about React 18 incompatibility:
```
ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17.
```
This fixes this issue by removing testing-library/react-hooks as it is now not being maintained and instead it's methods are being developed inside of testing-library/react. 
Also with the upgrades, there were a couple errors happening in AutoComplete's test cases, so fixed those as well.


HideBa (approved):
LGTM 👍

rot1024 (approved):

PR #256, opened 2022-06-21
feat: add timeline component

keiya01:
# Overview

## What I've done

I added Timeline component but only display memory ui.

## What I haven't done

- implement event like `auto play`, `speed up`, `moving pointer on memory`
- implement zoom feature

## How I tested

- Please see `atoms/Timeline` component on storybook.

## Screenshot

**As of 2022/06/21**

Screen Shot 2022-06-21 at 23 18 48

**Completed version**

Screen Shot 2022-06-24 at 21 43 20

PR #264, opened 2022-07-04
Feature/add timeline interaction

keiya01:
# Overview

## What I've done

I added Timeline component in https://github.com/reearth/reearth-web/pull/256.
And in this PR, I added interactions to Timeline component.

## What I haven't done

- Set some interaction buttons like play button, stop button.

## How I tested

- Move knob by click, drag.
- Zoom timeline.

## Screenshot

https://user-images.githubusercontent.com/34934510/177150926-697fea3b-9818-40b0-aefd-872e25fc444a.mov

## Memo

I will add some buttons like play button, stop button, etc in another PR.

### DIFF
--- a/src/components/atoms/Timeline/hooks.ts
+++ b/src/components/atoms/Timeline/hooks.ts
@@ -1,0 +1,109 @@
+import {
+  MouseEvent,
+  MouseEventHandler,
+  useCallback,
+  useEffect,
+  useState,
+  WheelEventHandler,
+} from "react";
+
+import { BORDER_WIDTH, GAP_HORIZONTAL, MAX_ZOOM_RATIO, PADDING_HORIZONTAL } from "./constants";
+import { TimeEventHandler, Range } from "./types";
+
+const convertPositionToTime = (e: MouseEvent, { start, end }: Range, gapHorizontal: number) => {
+  const curTar = e.currentTarget;
+  const width = curTar.scrollWidth - (PADDING_HORIZONTAL + BORDER_WIDTH) * 2 - gapHorizontal;
+  const rect = curTar.getBoundingClientRect();
+  const clientX = e.clientX - rect.x;
+  const scrollX = curTar.scrollLeft;
+  const posX = clientX + scrollX - (PADDING_HORIZONTAL + BORDER_WIDTH);
+  const percent = posX / width;
+  const rangeDiff = end - start;
+  const sec = rangeDiff * percent;
+  return start + sec;
+};
+
+type InteractionOption = {
+  range: Range;
+  onClick?: TimeEventHandler;
+  onDrag?: TimeEventHandler;
+};
+
+export const useTimelineInteraction = ({
+  range: { start, end },
+  onClick,
+  onDrag,
+}: InteractionOption) => {
+  const [isMouseDown, setIsMouseDown] = useState(false);
+  const [zoom, setZoom] = useState(1);
+  const gapHorizontal = GAP_HORIZONTAL * (zoom - Math.trunc(zoom) + 1);
+  const handleOnMouseDown = useCallback(() => {
+    setIsMouseDown(true);
+  }, []);
+  const handleOnMouseUp = useCallback(() => {
+    setIsMouseDown(false);
+  }, []);
+  const handleOnMouseMove: MouseEventHandler = useCallback(
+    e => {
+      if (!isMouseDown || !onDrag || !e.target) {
+        return;
+      }
+
+      onDrag(convertPositionToTime(e, { start, end }, gapHorizontal));
+
+      const scrollThreshold = 30;
+      const scrollAmount = 20;
+      const clientX = e.clientX;
+      const curTar = e.currentTarget;
+      const clientWidth = curTar.clientWidth;
+
+      if (clientX > clientWidth - scrollThreshold) {
+        curTar.scroll(curTar.scrollLeft + scrollAmount, 0);
+      }
+      if (clientX < scrollThreshold) {
+        curTar.scroll(curTar.scrollLeft - scrollAmount, 0);
+      }
+    },
+    [onDrag, start, end, isMouseDown, gapHorizontal],
+  );
+
+  useEffect(() => {
+    window.addEventListener("mouseup", handleOnMouseUp, { passive: true });
+    return () => {
+      window.removeEventListener("mouseup", handleOnMouseUp);
+    };
+  }, [handleOnMouseUp]);
+
+  const handleOnClick: MouseEventHandler = useCallback(
+    e => {
+      if (!onClick) {
+        return;
+      }
+      onClick(convertPositionToTime(e, { start, end }, gapHorizontal));
+    },
+    [onClick, end, start, gapHorizontal],
+  );
+
+  const handleOnWheel: WheelEventHandler = useCallback(
+    e => {
+      const { deltaX, deltaY } = e;
+      const isHorizontal = Math.abs(deltaX) > 0 || Math.abs(deltaX) < 0;
+      if (isHorizontal) {
+        return;
+      }
+
+      setZoom(() => Math.min(Math.max(1, zoom + deltaY * -0.01), MAX_ZOOM_RATIO));
+    },
+    [zoom],
+  );
+
+  return {
+    zoom,
+    events: {
+      onMouseDown: handleOnMouseDown,
+      onMouseMove: handleOnMouseMove,
+      onClick: handleOnClick,
+      onWheel: handleOnWheel,
+    },
+  };
+};
--- a/src/components/atoms/Timeline/index.stories.tsx
+++ b/src/components/atoms/Timeline/index.stories.tsx
@@ -1,4 +1,5 @@
 import { Meta, Story } from "@storybook/react";
+import { useCallback, useState } from "react";
 
 import Icon from "../Icon";
 
@@ -29,3 +30,19 @@ export const DefaultRange: Story<Props> = () => (
     renderKnob={() => <Icon icon="ellipse" alt="ellipse" size={25} />}
   />
 );
+
+export const Movable: Story<Props> = () => {
+  const [currentTime, setCurrentTime] = useState(() => Date.now() + 3600000);
+  const onClick = useCallback(setCurrentTime, [setCurrentTime]);
+  const onDrag = useCallback(setCurrentTime, [setCurrentTime]);
+  return (
+    <Timeline
+      // Forward a hour
+      currentTime={currentTime}
+      knobSize={25}
+      renderKnob={() => <Icon icon="ellipse" alt="ellipse" size={25} />}
+      onClick={onClick}
+      onDrag={onDrag}
+    />
+  );
+};
--- a/src/components/atoms/Timeline/index.tsx
+++ b/src/components/atoms/Timeline/index.tsx
@@ -1,12 +1,24 @@
 import { memo, ReactElement, useMemo } from "react";
 
-import { metricsSizes, styled } from "@reearth/theme";
+import { styled } from "@reearth/theme";
 import { XXSRegular } from "@reearth/theme/fonts";
 
-type Range = {
-  start: number;
-  end: number;
-};
+import {
+  BORDER_WIDTH,
+  DAY_SECS,
+  EPOCH_SEC,
+  GAP_HORIZONTAL,
+  HOURS_SECS,
+  MAX_ZOOM_RATIO,
+  MEMORY_INTERVAL,
+  MEMORY_ZOOM_INTERVAL,
+  MINUTES_SEC,
+  NORMAL_MEMORY_WIDTH,
+  PADDING_HORIZONTAL,
+  STRONG_MEMORY_WIDTH,
+} from "./constants";
+import { useTimelineInteraction } from "./hooks";
+import { Range, TimeEventHandler } from "./types";
 
 export type Props = {
   /**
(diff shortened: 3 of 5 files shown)


src/components/atoms/Timeline/index.tsx
```diff
@@ -159,23 +204,31 @@ const MemoryList: React.FC<MemoryListProps> = memo(
...
   box-sizing: border-box;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
 `;
 
 const MemoryBox = styled.div`
-  border: 1px solid ${({ theme }) => theme.colors.publish.dark.icon.weak};
+  border: ${BORDER_WIDTH}px solid ${({ theme }) => theme.colors.publish.dark.icon.weak};
```
KaWaite, line 231: This might be a personal preference, so you don't have to do this, but I feel like it is easier to read like this:
```
border: ${({ theme }) => `${BORDER_WIDTH}px solid ${theme.colors.publish.dark.icon.weak}`};
```
That way, the structure of the border value is upheld and not broken up as much. What do you think?

src/components/atoms/Timeline/index.tsx
```diff
@@ -73,33 +84,57 @@ const Timeline: React.FC<Props> = memo(
...
+      memoryInterval,
+      strongMemoryHours,
+    ]);
 
     return (
       <Container>
-        <MemoryBox>
-          <MemoryList start={startDate} memoryCount={memoryCount} hoursCount={hoursCount} />
+        <MemoryBox {...events}>
+          <MemoryContainer style={{ gap: `0 ${gapHorizontal}px` }}>
```
KaWaite, line 128: If there wasn't a specific reason to use inline styles, can you move the gap to MemoryContainer's style declaration below?
You might already know, but if you don't, you can pass `gapHorizontal` as a regular prop here, and then below in the styled component declaration write like so:
```
const MemoryContainer = styled.div<{gapHorizontal: number}>`.....`
```

rot1024 (commented):
- Memory means 目盛り? "Scale" is better I think if so.
- You can separate each components in `index.tsx` into multiple `.tsx` files: e.g.
   - index.tsx
   - MemoryList.tsx
- What is the difference between the hooks written in hooks.ts and index.tsx? Basically, I think it would be better to put the hooks together in hooks.ts.

src/components/atoms/Timeline/index.tsx
keiya01: `gapHorizontal` is too many changed so we should pass this to style props for improving performance.
https://emotion.sh/docs/best-practices#use-the-style-prop-for-dynamic-styles

src/components/atoms/Timeline/index.tsx
```diff
@@ -159,23 +204,31 @@ const MemoryList: React.FC<MemoryListProps> = memo(
           }
           return <Memory key={`${idx}`} isHour={isHour} isStrongMemory={isStrongMemory} />;
```
keiya01, line 205: This component is rendered so many times, so this may make performance degrade.
When we need to improve performance we can try [this section in emotion](https://emotion.sh/docs/performance).
But this will make readability degrade.

src/components/atoms/Timeline/index.tsx
keiya01: I fixed! [refactor: border style](https://github.com/reearth/reearth-web/pull/264/commits/d5d30090118b411b25aae0b9bacaf493b28a9678)

keiya01:
@rot1024 
I fixed your review.
- [chore: rename from memory to scale](https://github.com/reearth/reearth-web/pull/264/commits/082043d94d08e7403edf7d72c130c8f7810dad16)
- [refactor: seperate ScaleList from index.tsx](https://github.com/reearth/reearth-web/pull/264/commits/6133a14b6ac738e1a6b6f0e4c3a93d2be3668049)
- [refactor: move hooks to hooks.ts](https://github.com/reearth/reearth-web/pull/264/commits/2cf64e219a23fe6206665a068234c4345d53059a)

src/components/atoms/Timeline/index.tsx
```diff
@@ -18,6 +17,7 @@ import {
   STRONG_SCALE_WIDTH,
 } from "./constants";
 import { useTimelineInteraction } from "./hooks";
+import { ScaleList } from "./ScaleList";
```
rot1024, line 7: So you're of the school that doesn't use export default, it's nice to have the IDE complement working for you.

rot1024 (commented):
Very nice. And while you're at it, why don't you write unit tests for components, too, since react-testing-libraryis already set up?

src/components/atoms/Timeline/ScaleList.tsx
```diff
@@ -0,0 +1,117 @@
+import { memo } from "react";
+
+import { styled } from "@reearth/theme";
+import { XXSRegular } from "@reearth/theme/fonts";
```
KaWaite, line 4: Sorry 99%-100% of the time please use anything related to fonts through the Text atom. Then you can do like `<Text size="xxs">TEXT HERE</Text>`

src/components/atoms/Timeline/ScaleList.tsx
```diff
@@ -0,0 +1,117 @@
...
+  gapHorizontal: number;
+} & ScaleListInnerProps;
+
+export const ScaleList: React.FC<Props> = ({ gapHorizontal, ...rest }) => {
+  return (
+    <ScaleContainer style={{ gap: `0 ${gapHorizontal}px` }}>
+      <ScaleListInner {...rest} />
+    </ScaleContainer>
+  );
+};
```
KaWaite, line 18: Sorry if you haven't been given this yet. Typically I onboard front-end devs and would have given it to you, but for some reason I wasn't asked to for you. https://github.com/reearth/guides/blob/main/frontend/style.md#spread-and-rest-syntax
```suggestion
export const ScaleList: React.FC<Props> = ({ gapHorizontal, ...props }) => {
  return (
    <ScaleContainer style={{ gap: `0 ${gapHorizontal}px` }}>
      <ScaleListInner {...props} />
    </ScaleContainer>
  );
};
```

src/components/atoms/Timeline/ScaleList.tsx
```diff
@@ -0,0 +1,117 @@
...
+        if (isStrongScale && idx !== lastStrongScaleIdx) {
+          const d = new Date(start.getTime() + idx * EPOCH_SEC * scaleInterval);
+          const year = d.getFullYear();
+          const month = MONTH_LABEL_LIST[d.getMonth()];
+          const date = `${d.getDate()}`.padStart(2, "0");
+          const hour = `${d.getHours()}`.padStart(2, "0");
+          const label = `${month} ${date} ${year} ${hour}:00:00.0000`;
+
+          return (
+            <LabeledScale key={`${idx}`}>
```
KaWaite, line 66: Is there a reason you are writing it like `key={`${idx}`}` and not just `key={idx}`?

KaWaite: If you are converting it to a string, is there a reason you are?

src/components/atoms/Timeline/ScaleList.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const ScaleListInner: React.FC<ScaleListInnerProps> = memo(function ScaleListPresenter({
+  start,
+  scaleCount,
+  hoursCount,
+  scaleInterval,
+  strongScaleHours,
+}) {
+  const strongHours = hoursCount * strongScaleHours;
+  const lastStrongScaleIdx = strongHours * (scaleCount / strongHours);
```
KaWaite, line 51: Can you explain why you are calling things strongXXXX? It might be another translation hiccup, but not sure!

src/components/atoms/Timeline/hooks.ts
```diff
@@ -0,0 +1,213 @@
...
+    return () => {
+      window.removeEventListener("mouseup", handleOnMouseUp);
+    };
+  }, [handleOnMouseUp]);
+
+  const handleOnClick: MouseEventHandler = useCallback(
+    e => {
+      if (!onClick) {
+        return;
+      }
```
KaWaite, line 99: Keeps things more concise and doesn't really degrade readability with such a simple conditional
```suggestion
      if (!onClick) return;
```

src/components/atoms/Timeline/hooks.ts
```diff
@@ -0,0 +1,213 @@
...
+    [onClick, end, start, gapHorizontal],
+  );
+
+  const handleOnWheel: WheelEventHandler = useCallback(
+    e => {
+      const { deltaX, deltaY } = e;
+      const isHorizontal = Math.abs(deltaX) > 0 || Math.abs(deltaX) < 0;
+      if (isHorizontal) {
+        return;
+      }
```
KaWaite, line 111:
```suggestion
      if (isHorizontal) return;
```

src/components/atoms/Timeline/index.stories.tsx
```diff
@@ -29,3 +30,19 @@ export const DefaultRange: Story<Props> = () => (
     renderKnob={() => <Icon icon="ellipse" alt="ellipse" size={25} />}
   />
 );
+
+export const Movable: Story<Props> = () => {
+  const [currentTime, setCurrentTime] = useState(() => Date.now() + 3600000);
+  const onClick = useCallback(setCurrentTime, [setCurrentTime]);
+  const onDrag = useCallback(setCurrentTime, [setCurrentTime]);
```
KaWaite, line 37: Can you explain why you are wrapping the setCurrentTime in useCallbacks? Is it just for simpler typings as a Timeline prop? Or some other reason?

KaWaite: I know this is a stories file, so if this is just to make sure the example is doing something specific, and not how it'll be used actually inside Re:Earth, you can resolve this message!

src/components/atoms/Timeline/ScaleList.tsx
keiya01: `strong` means emphasized scale 🤔
I'm setting width to scale and normal scale should be `1px` or **strong** scale  should be `2px`.
https://github.com/reearth/reearth-web/blob/d5d30090118b411b25aae0b9bacaf493b28a9678/src/components/atoms/Timeline/ScaleList.tsx#L113-L114

src/components/atoms/Timeline/ScaleList.tsx
keiya01: 🙏 [refactor: use Text component](https://github.com/reearth/reearth-web/pull/264/commits/4c7da4da79cfbb23660b5ba4d7ea3edeb27e142f)

src/components/atoms/Timeline/ScaleList.tsx
keiya01: 👍 [chore: rename from rest to props](https://github.com/reearth/reearth-web/pull/264/commits/ecc9f7d2a9f59c3573dc53c5e2fc2465fbc8e81c)

src/components/atoms/Timeline/ScaleList.tsx
keiya01: This is because, I often use [this jsx-key eslint rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-key.md) but in reearth this rule is disabled so I fixed to use string.
[refactor: use number idx as key](https://github.com/reearth/reearth-web/pull/264/commits/400fd05e820310184d4c5f4564fa52904059f338)

keiya01: But maybe we can try to enable this eslint rule.

keiya01:
@rot1024 @KaWaite 
I added test [test: add some test](https://github.com/reearth/reearth-web/pull/264/commits/68615e0c341a4530d54215bb51245bec66803787)

src/components/atoms/Timeline/ScaleList.tsx
KaWaite: I am still a little confused what is what from the naming, which might be my lack of understanding for everything that goes in to a geographic scale. For now its fine, but as the component comes closer to completion, especially when I can use it in the PR preview, there is a chance I might ask for the naming to be changed to be clearer!

KaWaite (approved):

src/components/atoms/Timeline/ScaleList.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    "Jul 03 2022 15:00:00.0000",
+    "Jul 03 2022 18:00:00.0000",
+    "Jul 03 2022 21:00:00.0000",
+    "Jul 04 2022 00:00:00.0000",
+    "Jul 04 2022 03:00:00.0000",
+    "Jul 04 2022 06:00:00.0000",
+    "Jul 04 2022 09:00:00.0000",
+  ];
+
+  expectedStrongScaleLabelList.map(label => screen.getByText(label));
```
rot1024, line 77:
```suggestion
  expectedStrongScaleLabelList.forEach(label => {
    expect(screen.getByText(label)).toBeInTheDocument();
  });
```

- `Array.prototype.map` should not be used for side effects
- Add assertions to clarify your intent even if `getByText` emits errors when there is no match

src/components/atoms/Timeline/ScaleList.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    "Jul 04 2022 03:00:00.0000",
+    "Jul 04 2022 04:00:00.0000",
+    "Jul 04 2022 05:00:00.0000",
+    "Jul 04 2022 06:00:00.0000",
+    "Jul 04 2022 07:00:00.0000",
+    "Jul 04 2022 08:00:00.0000",
+    "Jul 04 2022 09:00:00.0000",
+  ];
+
+  expectedStrongScaleLabelList.map(label => screen.getByText(label));
```
rot1024, line 128:
```suggestion
  expectedStrongScaleLabelList.forEach(label => {
  	expect(screen.getByText(label)).toBeInTheDocument();
  });
```

rot1024 (changes requested):

src/components/atoms/Timeline/ScaleList.test.tsx
keiya01: I fixed!
[test: fix test format](https://github.com/reearth/reearth-web/pull/264/commits/43dd586a556bad7e71981f1f8c01c700ab146918)

rot1024 (approved):

PR #265, opened 2022-07-05
fix: layer drag and drop does not work with indicators

nina992:
# Overview
Fix the Dnd in selection indicator in cesium viewer.
 
## What I've done
Instead of disabling the selection indicator in the viewer constructor , I modified the element style itself , and by this way no need to reconstruct the viewer each time we change the indicator and also the Dnd should work fine.

### DIFF
--- a/src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
+++ b/src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
@@ -27,6 +27,18 @@ export default function Indicator({ className, property }: Props): JSX.Element |
   const [img, w, h] = useIcon({ image: indicator_image, imageSize: indicator_image_scale });
 
   useEffect(() => {
+    !(
+      property?.indicator?.indicator_type === "default" ||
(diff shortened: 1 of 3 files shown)


src/components/molecules/Visualizer/Engine/Cesium/core/Indicator.tsx
```diff
@@ -26,6 +26,18 @@ export default function Indicator({ className, property }: Props): JSX.Element |
   const { indicator_type, indicator_image, indicator_image_scale } = property?.indicator ?? {};
   const [img, w, h] = useIcon({ image: indicator_image, imageSize: indicator_image_scale });
 
+  useEffect(() => {
+    !(
+      property?.indicator?.indicator_type === "default" ||
+      property?.indicator?.indicator_type === undefined
```
rot1024, line 32:
```suggestion
      !indicator_type || indicator_type === "default"
```

rot1024 (changes requested):

rot1024 (approved):

PR #266, opened 2022-07-08
fix: unselect layer not work properly

airslice:
# Overview

When click on blank space of map it should unselect layers but it doesn't work due to a bug.

The bug comes from updateing selectLayer functions of organisms `CanvasArea` and molecusles `Visualizer`. In the past the `CanvasArea` select funtion will check if the layer can be got by `layers.findById` and after update this condition should be moved into `Visualizer` selectLayer function. When reviewing, the code looks like can be refactored and i did it without thinking clearly. My mistake.  

## What I've done

Update the `selectLayer` function.

## How I tested

Select layer and click on blank space to unselect it.


KaWaite (approved):
OK!

rot1024 (approved):

PR #267, opened 2022-07-08
refactor: refactor the infinite scroll wrapper and queries

nina992:
# Overview
The changes: 
1- Refactoring the queries of graphql  and remove nodes 
2- Modify all the queries on hooks 
3- Create (infinite scroll) wrapper atom and reuse it wherever the need for infinite scroll

## What I've done
The pages affected by this changes : 
1- Dashboard page 
2- setting pages : 
	1- Project list page 
	2- Assets page
	3- Datasets page
	4- setting page for each page 
	5- accessing pages from reearth-editor

### DIFF
--- a/src/components/atoms/InfiniteScroll/hooks.ts
+++ b/src/components/atoms/InfiniteScroll/hooks.ts
@@ -1,0 +1,23 @@
+import { MutableRefObject } from "react";
+
+export default (onLoadMore?: () => void) => {
+  const handleAutoFillPage = (ref: MutableRefObject<HTMLDivElement | undefined | null>) => {
+    if (ref.current && ref.current?.scrollHeight <= document.documentElement.clientHeight) {
+      onLoadMore?.();
+    }
+  };
+
+  const handleScrollToBottom = (
+    { currentTarget }: React.UIEvent<HTMLDivElement, UIEvent>,
+    threshold = 5,
+  ) => {
+    if (
+      currentTarget.scrollTop + currentTarget.clientHeight >=
+      currentTarget.scrollHeight - threshold
+    ) {
+      onLoadMore?.();
+    }
+  };
+
+  return { handleAutoFillPage, handleScrollToBottom };
+};
--- a/src/components/atoms/InfiniteScroll/index.tsx
+++ b/src/components/atoms/InfiniteScroll/index.tsx
@@ -1,0 +1,44 @@
+import React, { ReactNode, useEffect, useRef } from "react";
+
+import { styled } from "@reearth/theme";
+
+import useHooks from "./hooks";
+
+type Props = {
+  className?: string;
+  loading?: boolean;
+  hasMoreItems?: boolean;
+  children?: ReactNode;
+  onGetMoreItems?: () => void;
+};
+
+const InfiniteScrollWrapper: React.FC<Props> = ({
+  className,
+  loading,
+  hasMoreItems,
+  onGetMoreItems,
+  children,
+}) => {
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const { handleAutoFillPage, handleScrollToBottom } = useHooks(onGetMoreItems);
+
+  useEffect(() => {
+    if (wrapperRef.current && !loading && hasMoreItems) handleAutoFillPage(wrapperRef);
+  }, [handleAutoFillPage, hasMoreItems, loading, onGetMoreItems]);
+
+  return (
+    <Wrapper
+      ref={wrapperRef}
+      className={className}
+      onScroll={e => !loading && hasMoreItems && handleScrollToBottom(e)}>
+      {children}
+    </Wrapper>
+  );
+};
+const Wrapper = styled.div`
+  width: 100%;
+  height: 100%;
+  overflow-y: scroll;
+  -webkit-overflow-scrolling: touch;
+`;
+export default React.memo(InfiniteScrollWrapper);
--- a/src/components/organisms/Common/AssetContainer/hooks.ts
+++ b/src/components/organisms/Common/AssetContainer/hooks.ts
@@ -2,7 +2,6 @@ import { useApolloClient } from "@apollo/client";
 import { useCallback, useState, useEffect } from "react";
 
 import {
-  GetAssetsQuery,
   useCreateAssetMutation,
   useRemoveAssetMutation,
   useGetAssetsQuery,
@@ -11,8 +10,6 @@ import {
 } from "@reearth/gql";
 import { useT } from "@reearth/i18n";
 import { useNotification } from "@reearth/state";
-
-export type AssetNodes = NonNullable<GetAssetsQuery["assets"]["nodes"][number]>[];
 
 export type AssetSortType = "date" | "name" | "size";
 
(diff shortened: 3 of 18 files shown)


src/components/atoms/InfiniteScroll/hooks.ts
```diff
@@ -0,0 +1,23 @@
+import { MutableRefObject } from "react";
+
+export default (onLoadMore?: () => void) => {
+  const handleAutoFillPage = (ref: MutableRefObject<HTMLDivElement | undefined | null>) => {
+    if (ref.current && ref.current?.scrollHeight <= document.documentElement.clientHeight) {
+      onLoadMore?.();
+    }
```
KaWaite, line 7: I think something about this isn't quite working correctly. In the Asset settings page you get this called when the page loads, before any scrolling. It isn't a BIG deal, since it only seems to call once, but it really shouldn't call UNLESS the page is too long for the initial fetch.

src/components/organisms/Common/AssetContainer/hooks.ts
```diff
@@ -74,7 +71,21 @@ export default (teamId?: string, initialAssetUrl?: string | null, allowDeletion?
...
+            id: asset.node.id,
+            name: asset.node.name,
+            size: asset.node.size,
+            teamId: asset.node.teamId,
+            url: asset.node.url,
+            contentType: asset.node.contentType,
+          }
+        : undefined,
+    )
+    .filter((asset): asset is Asset => !!asset);
```
KaWaite, line 88: I think you can keep it more concise if you do something like this:
```
const assets = data?.assets.edges
    .map<Asset | undefined>(({ node })=> node)
    .filter((asset): asset is Asset => !!asset);
```

Please test to make sure it works though.
Also, what is the purpose of the filter? Is there any chance that there'll be an undefined asset inside the assets array?

src/components/atoms/InfiniteScroll/hooks.ts
airslice: Just happen to see this. I think this part comes from my suggestion some weeks ago. This is only for solving large screen issue (first fetch cannot fill the area) and yes it should be called only under this situation but the point is it seems to be some difficult to check if it is too long (from my memory it was due to the DOM structure is different with different usecases and the function is a common util at that time). So my idea was always want the content fill the area and if not and if there's more data then load more. At that time i only look into this with load projects in large screen issue and i'm not sure about other situations. And also i think the code changes a lot now maybe can optimize this with adding an area's ref?

src/components/organisms/Dashboard/hooks.ts
```diff
@@ -117,25 +114,23 @@ export default (teamId?: string) => {
...
+              image: project.node.imageUrl,
+              status: toPublishmentStatus(project.node.publishmentStatus),
+              isArchived: project.node.isArchived,
+              sceneId: project.node.scene?.id,
             }
           : undefined,
       )
       .filter((project): project is Project => !!project);
-  }, [projectNodes]);
+  }, [projectData?.projects.edges]);
```
KaWaite, line 133: Similar to the comment above, you can destructure the node so you can clean up the assignment below, something like this:

```
 const projects = useMemo(() => {
    return (projectData?.projects.edges ?? [])
      .map<Project | undefined>(({node})=>
           node
          ? {
              id: node.id,
              description: node.description,
              name: node.name,
              image: node.imageUrl,
              status: toPublishmentStatus(node.publishmentStatus),
              isArchived: node.isArchived,
              sceneId: node.scene?.id,
            }
          : undefined,
      )
      .filter((project): project is Project => !!project);
  }, [projectData?.projects.edges]);
```
You can also destructure further like this:
```
(projectData?.projects.edges ?? [])
      .map<Project | undefined>(({node: { id, description, name, imageUrl: image, publishmentStatus: status, isArchived, scene: {id?: sceneId}}})=> {
   id,
   description,
   name,
   etc, 
   etc,
}
```
The second one might be too much? Also, you can go even further to ONLLY destructure specifically the keys that are different between the data and the frontend's Project (aka image VS imageUrl, etc) and leave the rest as `...rest`. But again, it might be too much and the code might become harder to understand. I'll let you play around and find whats best here

src/components/organisms/Settings/Project/Dataset/hooks.ts
```diff
@@ -38,9 +34,11 @@ export default (projectId: string) => {
...
   });
 
-  const nodes = data?.datasetSchemas.edges.map(e => e.node) ?? [];
-
-  const datasetSchemas = nodes.filter(Boolean) as DatasetSchemas;
+  const datasetSchemas = data?.datasetSchemas.edges
+    .map<Item | undefined>(item =>
+      item.node ? { id: item.node.id, name: item.node.name } : undefined,
+    )
+    .filter((item): item is Item => !!item);
```
KaWaite, line 41: Same as above: destructure the node and then you don't need to assign the values to keys. And again, not sure if the filter is necessary? Let me know

src/components/organisms/Settings/Project/Public/hooks.ts
```diff
@@ -42,26 +42,27 @@ export default ({ projectId }: Params) => {
...
   const rawProject = useMemo(
-    () => data?.projects.nodes.find(p => p?.id === projectId),
+    () => data?.projects.edges.find(p => p.node?.id === projectId),
     [data, projectId],
   );
+
   const project = useMemo(
     () =>
-      rawProject?.id
+      rawProject?.node?.id
```
KaWaite, line 51: Destructure node like above

src/components/organisms/Settings/Project/hooks.ts
```diff
@@ -26,25 +26,26 @@ export default ({ projectId }: Params) => {
...
   const rawProject = useMemo(
-    () => data?.projects.nodes.find(p => p?.id === projectId),
+    () => data?.projects.edges.find(p => p.node?.id === projectId),
     [data, projectId],
   );
+
   const project = useMemo(
     () =>
-      rawProject?.id
+      rawProject?.node?.id
```
KaWaite, line 35: Destructure node

src/components/organisms/Settings/ProjectList/hooks.ts
```diff
@@ -65,19 +62,17 @@ export default (teamId: string) => {
     }
   }, [currentTeam, team, setTeam]);
 
-  const projectNodes = projectData?.projects.edges.map(e => e.node) as ProjectNodes;
-
-  const currentProjects = (projectNodes ?? [])
+  const currentProjects = projectData?.projects.edges
     .map<Project | undefined>(project =>
-      project
+      project.node
```
KaWaite, line 67: Destructure node

src/components/atoms/InfiniteScroll/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  hasMoreItems,
+  onGetMoreItems,
+  children,
+}) => {
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const { handleAutoFillPage, handleScrollToBottom } = useHooks(onGetMoreItems);
+
+  useEffect(() => {
+    if (wrapperRef.current && !loading && hasMoreItems) handleAutoFillPage(wrapperRef);
+  }, [handleAutoFillPage, hasMoreItems, loading, onGetMoreItems]);
```
KaWaite, line 27: This component doesn't have too many hooks, so you don't really need the hooks.ts file. But if you want to keep the file, please bring the ref and useEffect into the hooks file as well to keep things consistent (unless there is a reason you separated them, in which case please let me know).

KaWaite (commented):
Looks really good! A few clean up comments but really good refactor!

PR #277, opened 2022-07-16
feat: add timeline playing buttons

keiya01:
# Overview

## What I've done

I added feature for playing timeline.
And I added feature for changing color.

## What I haven't done

- Finally timeline component is completed.

## How I tested

- [ ] Click play button and check if knob is forwarding.
- [ ] Click playback button and check if knob is backing.
- [ ] Change speed and check if knob is moving as expected.

## Screenshot

https://user-images.githubusercontent.com/34934510/179349463-8724d23b-f06b-4dda-87ca-e239f56f8fb2.mov


rot1024 (approved):
LGTM

src/components/atoms/Timeline/index.tsx
```diff
@@ -35,15 +52,76 @@ const Timeline: React.FC<Props> = memo(
       strongScaleHours,
       currentPosition,
       events,
+      player: {
+        playSpeed,
+        onChangePlaySpeed,
```
KaWaite, line 57: onPlaySpeedChange

Please see [here](https://github.com/reearth/guides/blob/main/frontend/style.md#functions) for word order of functions

src/components/atoms/Timeline/index.tsx
```diff
@@ -35,15 +52,76 @@ const Timeline: React.FC<Props> = memo(
...
+        formattedCurrentTime,
+        isPlayed,
+        isPlayBacked,
+        toggleIsPlayed,
+        toggleIsPlayBacked,
```
KaWaite, line 62: Are `isPlayBacked` and `toggleIsPlayBacked` about playing in reverse order? If so, perhaps its better to be `isPlaying`, `isPlayingReversed`, `toggleIsPlaying` and `toggleIsPlayingReversed`.
If you can think of a better way to name them, or I am not quite right about what these mean, please let me know!

src/components/atoms/Timeline/index.tsx
```diff
@@ -35,15 +52,76 @@ const Timeline: React.FC<Props> = memo(
...
     });
+    const t = useT();
+
+    if (!isOpened) {
+      return (
+        <OpenButton onClick={onOpen}>
+          <Icon alt={t("Open timeline")} icon="timeline" size={24} />
+        </OpenButton>
+      );
+    }
```
KaWaite, line 79: In Re:Earth, usually when we have conditional rendering we will just do a ternary operator in the main return.
In this case it'd be like
```
return isOpened ? (
<Container>
...
...
</Container>) : (
   <OpenButton onClick={onOpen}>
          <Icon alt={t("Open timeline")} icon="timeline" size={24} />
   </OpenButton>
)
```

src/components/atoms/Timeline/index.tsx
```diff
@@ -35,15 +52,76 @@ const Timeline: React.FC<Props> = memo(
...
 
     return (
       <Container>
+        <CloseButton themeColor={themeColor} onClick={onClose}>
+          <Icon alt={t("Close timeline")} icon="cancel" size={16} />
+        </CloseButton>
+        <ToolBox>
+          <>
```
KaWaite, line 87: I can't see a reason for a fragment here, so if it isn't needed please remove it.

src/components/atoms/Timeline/index.tsx
```diff
@@ -35,15 +52,76 @@ const Timeline: React.FC<Props> = memo(
...
+                isRight
+                themeColor={themeColor}
+                isPlayed={isPlayed}
+                onClick={toggleIsPlayed}>
+                <Icon alt={t("Play timeline")} icon="playRight" size={16} />
+              </PlayButton>
+            </li>
+            <li>
+              <InputRangeLabel>
+                <>
```
KaWaite, line 107: Same here. Please remove if not needed

src/components/atoms/Timeline/index.tsx
```diff
@@ -22,10 +24,25 @@ export type Props = {
...
+  function TimelinePresenter({
+    currentTime,
+    range,
+    onClick,
+    onDrag,
+    onPlay,
+    isOpened,
+    onOpen,
+    onClose,
+    themeColor,
```
KaWaite, line 44: Why do you need a themeColor passed in here? If this is to support the published theme, please refer to the other builtin widgets (Storytelling or Button) for full implementation, but roughly it uses a helper function usePublishTheme and you need to pass the sceneProperty to the component.
```
const publishedTheme = usePublishTheme(sceneProperty?.theme);
```
If this timeline will become an ACTUAL widget, and will be passed in to the Widget component like Storytelling and Button, please also reference that component to see what props you already get with it!

src/i18n/translations/en.yml
```diff
@@ -5,6 +5,11 @@ Notfound: Notfound
 Published: Published
 Unpublished: Unpublished
 not set: not set
+Open timeline: 'Open timeline'
+Close timeline: 'Close timeline'
+Playback timeline: 'Playback timeline'
+Play timeline: 'Play timeline'
+ellipse: 'ellipse'
```
KaWaite, line 12: I am not familiar with screen readers and how they are used enough to know, but are these detailed enough? Will aria labels typically be short?

src/components/atoms/Timeline/hooks.ts
```diff
@@ -119,6 +122,91 @@ const useTimelineInteraction = ({
   };
 };
 
+const DEFAULT_INTERVAL = 10;
```
KaWaite, line 125: This can go into constants, right?

src/i18n/translations/en.yml
keiya01: Typically `aria-label` should be set text same as able-bodied person see usually.
If UI has button like `<button><Icon /></button>`, we can understand what button means from icon but visually impaired person can not see icon so this person can not understand about this button. So we need to describe about Icon by `aria-label`.
If we set aria-label like `Open timeline`, screen reader says `"Open timeline button is clicked"` by clicking `<button><Icon /></button>`.

src/components/atoms/Timeline/index.tsx
keiya01: This is fixed by [chore: rename to onPlaySpeedChange](https://github.com/reearth/reearth-web/pull/277/commits/8848e5506830ecceb5b4c1bc83eb4d7600d703a5)

src/components/atoms/Timeline/index.tsx
keiya01: This is fixed by [chore: rename playing boolean vars](https://github.com/reearth/reearth-web/pull/277/commits/7e73fb2aa13090f87552bcaaa04fe4d00e5e6969)

src/components/atoms/Timeline/index.tsx
keiya01: This was fixed by [refactor: use conditional operator](https://github.com/reearth/reearth-web/pull/277/commits/f9e3a78b36e4bf9a7c785367a43b05592e23d794)

src/components/atoms/Timeline/index.tsx
keiya01: This was fixed by [refactor: remove unnecessary fragment](https://github.com/reearth/reearth-web/pull/277/commits/f74dd9579b4db1d4ad4456f8e8afca3f149cf007)

src/components/atoms/Timeline/hooks.ts
keiya01: This was fixed by [refactor: move default interval constant](https://github.com/reearth/reearth-web/pull/277/commits/8dc07898743fe1379e772521ece40247bc591914)

src/components/atoms/Timeline/index.tsx
keiya01: This is fixed by [fix: use publishTheme](https://github.com/reearth/reearth-web/pull/277/commits/ba9810cc9acb32d842c134d6ce2253008c2a1824)

src/components/atoms/Timeline/index.tsx
```diff
@@ -22,10 +24,25 @@ export type Props = {
   range?: { [K in keyof Range]?: Range[K] };
   onClick?: TimeEventHandler;
   onDrag?: TimeEventHandler;
+  onPlay?: TimeEventHandler;
+  onOpen?: () => void;
+  onClose?: () => void;
+  isOpened?: boolean;
+  publishedTheme?: PublishTheme;
```
KaWaite, line 31: I am not sure really the plan for HOW we are going to use this component. If it'll be different from the builtin Widgets then perhaps this is fine. But if it'll be treated like a builtin widget this will have to be replaced with sceneProperty and then the PublishedTheme will be retrieved from here, just like the Storybook, Button, etc.
For now lets leave this as is since maybe that hasn't been decided yet.

KaWaite (approved):
OK!

src/i18n/translations/en.yml
KaWaite: Ah ok! So these terms will be used in a sentence by the screenreader. Ok!

src/components/atoms/Timeline/index.tsx
keiya01: Timeline component will be used as builtin widget. So I fixed.
[fix: use sceneProperty](https://github.com/reearth/reearth-web/pull/277/commits/c4dd273590f4df91bad30bc33c2a89b101be9369)

src/components/atoms/Timeline/index.tsx
```diff
@@ -1,8 +1,11 @@
 import { memo } from "react";
 
-import { styled } from "@reearth/theme";
+import { useT } from "@reearth/i18n";
+import { PublishTheme, styled, usePublishTheme } from "@reearth/theme";
 
+import { SceneProperty } from "../../molecules/Visualizer";
```
KaWaite, line 6: Please use `@reearth/components` for URLs that aren't heavily nested AND tightly related. (example: its fine that ScaleList.tsx is imported relatively)
While you are at it please update the other atoms' imports as well.

keiya01: This is fixed by [chore: update some import format](https://github.com/reearth/reearth-web/pull/277/commits/fa94d7008f59a430ecdc9db46bf2a464d98a1b0b)

KaWaite (approved):

PR #279, opened 2022-07-21
fix: pointer events issues around widgets

KaWaite:
## What I've done
fixed:
- Not being able to scroll beside smaller widgets when a bigger widget is in the same GridArea
- Being able to use widgets (ie. click on buttons) while in WAS editor mode. If you can resize, the overlay disappeared, and then you could use the widget. Now, no pointerEvents happen inside Widgets when in WAS editor mode.

## How I tested
- Manually tested all builtin widgets, as well as a plugin widget.


src/components/molecules/Visualizer/Widget/index.tsx
```diff
@@ -84,7 +84,9 @@ export default function WidgetComponent<PP = any, SP = any>({
     : "both";
 
   return Builtin ? (
-    <Builtin {...props} widget={w} layout={layout} extended={extended} onExtend={onExtend} />
+    <BuiltinWrapper editing={editing}>
+      <Builtin {...props} widget={w} layout={layout} extended={extended} onExtend={onExtend} />
+    </BuiltinWrapper>
```
rot1024, line 89: It does not seem to be enough to require a separate component.

```suggestion
    <div style={{ pointerEvents: editing ? "none" : "auto" }}>
      <Builtin {...props} widget={w} layout={layout} extended={extended} onExtend={onExtend} />
    </div>
```

rot1024 (approved):

PR #280, opened 2022-07-22
feat: add mouse events to plugin API

airslice:
close https://github.com/reearth/reearth/issues/299
close https://github.com/reearth/reearth/issues/298

# Overview

This feature extends plugin API event with multiple mouse events:

### Events:

On mouse keys:
- click
- doubleclick
- mousedown
- mouseup
- rightclick
- rightdown
- rightup
- middleclick
- middledown
- middleup

On mouse behavior:
- mousemove
- mouseenter
- mouseleave

On mouse wheel:
- wheel

Also conatins:
- pinchstart
- pinchend
- pinchmove

### Event Props:

- wheel: `{ delta: number; }`
- others: `{
  x?: number;
  y?: number;
  lat?: number;
  lng?: number;
  height?: number;
  layerId?: string;
}`

The implementation for engine `cesium` is base on `resium` `Viewer` events.

## What I've done

- Extend engineRef methods with mouse events.
- Implement mouse events with engine `cesium` by using `resium` events.
- Pass the engine method to plugin content through `providerProps`.
- Call engine method inside plugin content to bind mouse events with plugin event `emit`.
- Extend plugin event types.

## What I haven't done

Known issue:
- The plugin event `off` doesn't work well when calling from plugin. It might be related with quickjs and will not be handled  in this PR.

## How I tested

- Call plugin API to add event callbacks from both console and plugin.
- Here's [a simple test plugin](https://github.com/airslice/reearth-plugin-api-demo/blob/main/packages/reearth-plugin-api-demo-0.0.2.zip).

## Screenshot

image

Known issue:
- The implementation of using `resium` `Viewer` events is somehow tortuous and the code added in engineRef also got some affected. I use a middle variable `mouseEventCallbacks` to store the callbacks and export it in order to pass it to resium `Viewer` component. This will perhaps lead to some problem when we add other engines in the future (not much maybe).
- The implementation of using plugin event (binding `emit` to events) will lead to some perfermance waste (but not much).


rot1024 (changes requested):

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -484,6 +485,13 @@ function useProviderProps(
     [engineRef],
   );
 
+  const onMouseEvent = useCallback(
+    (eventType: keyof EngineRef, fn: any) => {
```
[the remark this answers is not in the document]
airslice: Updated.

src/components/molecules/Visualizer/Plugin/types.ts
```diff
@@ -32,12 +32,37 @@ export type Reearth = {
   ) => void;
 };
 
+export type MouseEventProps = {
```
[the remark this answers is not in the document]
airslice: Updated.

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -4,7 +4,59 @@ import { LatLngHeight, Camera, Typography, Rect } from "@reearth/util/value";
...
+  y?: number;
+  lat?: number;
+  lng?: number;
+  height?: number;
+  layerId?: string;
+  delta?: number;
+};
+
+export type MouseEvents = {
+  click: ((props: MouseEventProps) => void) | undefined;
```
KaWaite, line 18: Can you choose in a single file either `something?: someType` OR `something: someType | undefined` and stick with it? MouseEventProps uses the first way and then the rest use the second way.

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -173,8 +180,67 @@ export default ({
...
+          ...(position ? getLocationFromScreenXY(viewer.scene, position.x, position.y) ?? {} : {}),
+        };
+        const layerId = getLayerId(target);
+        if (layerId) props.layerId = layerId;
+        engineAPI.mouseEventCallbacks[type]?.(props);
+      }
+    },
+    [engineAPI],
+  );
+  const handleMouseWheel = useCallback(
```
KaWaite, line 201:
```suggestion
  );
  
  const handleMouseWheel = useCallback(
```

src/components/molecules/Visualizer/Engine/ref.ts
airslice: Sometimes when i want to express one property is optional i use `something?: somtType` and when i want to express some property exists there but its value can be `undefined` i use `something?: someType | undefined`. But yes these two should be equal and i'll update to use the first one maybe.

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
airslice: updated.

src/components/molecules/Visualizer/Engine/ref.ts
airslice: Sorry i changed this part back as i found there's some difference between the two syntax. When it comes to use `keyof` with the type it will got different result and has effects on other type definitions.

src/components/molecules/Visualizer/Plugin/types.ts
```diff
@@ -32,12 +32,39 @@ export type Reearth = {
...
```
rot1024, line 35:
```suggestion
export type MouseEvent = {
```

Props is a term of React. But here is not React.

rot1024 (approved):

PR #284, opened 2022-07-25
test: introduce vitest

rot1024:
- Use Vitest and get ultra fast tests
   - Fix problem that jest-yaml-transform reports an error that node version of engine field in package.json is not met
   - Use react-inlinesvg instead of react-svg to avoid errors with Vitest
- Refactor some unit tests
- Add role prop to Flex and Icon component to improve accessibility


src/components/atoms/Flex/index.tsx
```diff
@@ -33,21 +35,28 @@ const Flex: React.FC<Props> = ({
...
+        flexWrap: wrap,
+        flexBasis: basis,
+        flexGrow: grow,
+        flexShrink: shrink,
+        flex,
+        gap, // TODO: Safari doesn't support this property and please develop polyfill
+      }}
+      onClick={onClick}
+      data-testid={testId}
+      aria-hidden={!hidden}>
```
KaWaite, line 59: Why is it negated? Shouldn't `aria-hidden` = true if `hidden` is true? 
Instead wouldn't it make more sense to make `hidden`'s default true. That way, devs would have to specifically set the Flex to false so that screen readers know it is interactive. Unless the other way makes more sense and unless stated ALL Flex components are considered interactive.

src/components/atoms/Flex/index.tsx
```diff
@@ -1,10 +1,12 @@
-import { ReactNode, CSSProperties } from "react";
+import { ReactNode, CSSProperties, AriaRole } from "react";
 
 export type Props = {
   className?: string;
   onClick?: () => void;
   children?: ReactNode;
   testId?: string;
+  role?: AriaRole;
+  hidden?: boolean;
```
KaWaite, line 9: Should these be `ariaRole` and `ariaHidden`? I'm not sure these are clear enough as is. Especially hidden sounds like I can hide the Flex component and its contents by passing true to there.

src/components/atoms/Flex/index.tsx
rot1024: Fixed: [`5359564` (#284)](https://github.com/reearth/reearth-web/pull/284/commits/5359564bad4035b2cd6c8805106262c713484aa4)

KaWaite (approved):

PR #285, opened 2022-07-28
feat: add builtin timeline widget

keiya01:
# Overview

## What I've done

- Connected timeline component with cesium.
- Displayed timeline component as builtin widget.

## How I tested

- run reearth-backend with https://github.com/reearth/reearth-backend/pull/161
- run reearth-web
- then you can add timeline widget from side panel
- and you can operate timeline component

## Screenshot

https://user-images.githubusercontent.com/34934510/181517884-fb918165-d0fb-44ae-a7b0-d9aebc363924.mov


src/components/molecules/Visualizer/hooks.ts
```diff
@@ -65,6 +69,33 @@ export default ({
...
+          if (!widget) {
+            return false;
+          }
+          return !!widget.widgets?.some(
+            w => `${w.pluginId}/${w.extensionId}` === TIMELINE_BUILTIN_WIDGET_ID,
+          );
+        });
+      });
+    });
+  }, [alignSystem]);
```
keiya01, line 97: I want to check if timeline widget is displayed but is there a smarter way to do this?
This is used for `requestRenderMode` and `maximumRenderTimeChange` props in Cesium component.
This is because, if we don't set these props and we operate timeline widget then cesium does not render earth by operating timeline widget.

src/util/value.ts
```diff
@@ -24,6 +24,8 @@ export type Camera = {
   fov: number;
 };
 
+export type Clock = CesiumClock | undefined;
```
KaWaite, line 27: This file is typically for value types and easy conversion between typescript types and GQL value types (please see below code, especially the comment stating any new field should be added to valueTypeMapper). Unless I am mistaken, perhaps you should move this type somewhere else! If it is needed for converting between GQL types as well, please follow the comment's instructions

src/components/molecules/Visualizer/hooks.ts
KaWaite: I'm sure @rot1024 has a better solution than what I'll suggest, but I think checking the widget [here](https://github.com/reearth/reearth-web/blob/7eead9b459f562b8b4c21f4cf571311d9c1c25af/src/components/organisms/EarthEditor/CanvasArea/convert.ts#L160) where we set the widgets you should do this. Do like a `let hasTimelineBuiltinWidget: boolean;` or useState and then inside the `convertWidgets` set it to true if there is a timeline widget

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -211,6 +242,7 @@ export default ({
     innerCamera,
     infobox,
     overriddenSceneProperty,
+    shouldRender: hasTimelineBuiltinWidget,
```
KaWaite, line 245: I'd like @rot1024's input since I am not fully sure on this, but I feel like this is very specific to Cesium and not sure it belongs here.

src/util/value.ts
KaWaite: Also, I think it is clearer to have this typed as `export type Clock = CesiumClock` and then where you use it state it can be `undefined`.

src/components/molecules/Visualizer/Widget/builtin.ts
```diff
@@ -2,14 +2,18 @@ import Button from "./Button";
...
 import type { Component } from ".";
 
+export const TIMELINE_BUILTIN_WIDGET_ID = "reearth/timeline";
+
 const builtin: Record<string, Component> = {
   "reearth/menu": Menu,
   "reearth/button": Button,
   "reearth/splashscreen": SplashScreen,
   "reearth/storytelling": Storytelling,
+  [TIMELINE_BUILTIN_WIDGET_ID]: Timeline,
```
KaWaite, line 16: If possible I'd like to keep things consistent here. Can you export the other builtin widget ids. We aren't using them currently anywhere else, but you never know and keeps the code here more predictable.

src/components/molecules/Visualizer/Widget/Timeline/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+      />
+    </Widget>
+  );
+};
+
+const Widget = styled.div<{
+  extended?: boolean;
+}>`
+  max-width: 100vw;
+  width: ${({ extended }) => (extended ? "100%" : "500px")};
```
KaWaite, line 29: Is the 500px based on the design? I'm guessing you are copying this from Storytelling, which is good because I think they are very similar widgets, but please double check with @lavalse for a good base width.

src/components/molecules/Visualizer/Widget/Timeline/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  @media (max-width: 560px) {
+    width: ${({ extended }) => (extended ? "100%" : "90vw")};
```
KaWaite, line 32: Here too, please check with @lavalse

KaWaite (commented):
Please see comments

src/components/molecules/Visualizer/hooks.ts
keiya01: I fixed your review in [refactor: add ownBuiltinWidgets property to convertWidgets](https://github.com/reearth/reearth-web/pull/285/commits/af54651e818dc7c529e103862cd5252646a59d61).
Thank you for your advice!!
And this commit fixes the following your comment.
- https://github.com/reearth/reearth-web/pull/285#discussion_r934156880
- https://github.com/reearth/reearth-web/pull/285#discussion_r934160698

src/components/molecules/Visualizer/Widget/builtin.ts
keiya01: This is fixed by [type: fix clock type](https://github.com/reearth/reearth-web/pull/285/commits/582f1c83d5a7479b2840fcb440a185ae0ebf131a).
Thank you:pray:

src/components/molecules/Visualizer/Widget/index.tsx
```diff
@@ -73,8 +73,8 @@ export default function WidgetComponent<PP = any, SP = any>({
 
   if (!w) return null;
 
-  const Builtin =
-    w?.pluginId && w.extensionId ? builtin[`${w.pluginId}/${w.extensionId}`] : undefined;
+  const builtinWidgetId = `${w.pluginId}/${w.extensionId}`;
```
KaWaite, line 76: This is't necessarily a built-in widget's ID. 
```suggestion
  const widgetId = `${w.pluginId}/${w.extensionId}`;
```

src/components/molecules/Visualizer/Widget/index.tsx
```diff
@@ -73,8 +73,8 @@ export default function WidgetComponent<PP = any, SP = any>({
...
+  const Builtin = isBuiltinWidget(builtinWidgetId) ? builtin[builtinWidgetId] : undefined;
```
KaWaite, line 77: Please tell me if you think I'm wrong (because I'm not overly confident in dealing with Records, but are we not overly complicating this (both the old code and the new code)?
Can't we just do
```
  const Builtin = builtin[`${w.pluginId}/${w.extensionId}`];
```
and then it'll either return a builtin component to render below or be undefined?

keiya01: The reason why I'm using `isBuiltinWidget()`  means that I want to assert `builtinWidgetId` as `builtin` object's key. If we don't assert `builtintWidgetId` then we can't access `builtin` object without `as`  keyword like the following.

```
builtin[builtinWidgetId as keyof typeof builtin];
```

But this is not reliable...

src/components/molecules/Visualizer/Widget/Timeline/index.tsx
keiya01: This was fixed by [style: fix width to 720px](https://github.com/reearth/reearth-web/pull/285/commits/8e45fb971b4f2f1dba37be89cbbc551455e48d54)

src/components/molecules/Visualizer/Widget/Timeline/index.tsx
keiya01: According to him there is no need to change this style

KaWaite (approved):

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -61,6 +62,7 @@ export type EngineRef = {
   zoomIn: (amount: number) => void;
   zoomOut: (amount: number) => void;
   changeSceneMode: (sceneMode: SceneMode | undefined, duration?: number) => void;
+  getClock: () => Clock | undefined;
```
rot1024, line 65: It has become dependent on Cesium, so it should be a generic type independent from the specific map engine.

e.g.

```ts
type Clock = {
  isPlaying: boolean;
  setPlaying: (playing: boolean) => void;
  currentTime: Date; // note that it should be JS Date
};
```

src/components/molecules/Visualizer/Plugin/context.tsx
```diff
@@ -40,6 +41,7 @@ export type Props = {
   sceneProperty?: any;
   tags?: Tag[];
   camera?: CameraPosition;
+  clock: () => Clock | undefined;
```
rot1024, line 44: Instead, define a Clock type that is independent of a specific map engine.

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/ref.ts
keiya01: I fixed this comment.
[fix: abstract clock value](https://github.com/reearth/reearth-web/pull/285/commits/8ef72f3e3075b397e1130d9042b037333cbc6690)

src/components/molecules/Visualizer/Plugin/types.ts
```diff
@@ -1,4 +1,4 @@
-import { Clock } from "cesium";
+import { Clock } from "../Engine/ref";
```
rot1024, line 1: Define Clock type here and import it in `Engine/Cesium/ref.ts` etc. since this file should be a master with all types explicitly defined for the plugin's API.

Current: Plugin/types ---use--> Engine/ref (Clock type)
Ideal: Plugin/types (Clock type) <---use-- Engine/ref

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -49,6 +48,19 @@ export type MouseEventHandles = {
...
+export type Clock = {
+  startTime: Date;
+  stopTime: Date;
+  currentTime: Date;
+  tick: () => Date;
+  shouldAnimate: boolean;
+  onTick: {
+    addEventListener: ClockEventHandler<() => void>;
+    removeEventListener: ClockEventHandler;
+  };
```
rot1024, line 61: To make the plugin API easier to understand, instead of exposing an event object, let's add a tick event to the plugin, similar to the cameramove event.

Add a new event "tick" to https://github.com/reearth/reearth-web/blob/main/src/components/molecules/Visualizer/Plugin/types.ts#L45 and implement rest

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/ref.ts
```diff
@@ -48,6 +48,19 @@ export type MouseEventHandles = {
   onWheel: (fn: MouseEvents["wheel"]) => void;
 };
 
+type ClockEventHandler<R = void> = (cb: (clock: Clock) => void) => R;
+export type Clock = {
+  startTime: Date;
+  stopTime: Date;
+  currentTime: Date;
+  tick: () => Date;
+  shouldAnimate: boolean;
```
rot1024, line 57:
```suggestion
  isPlaying: boolean;
```

I think it's easy to understand

rot1024 (changes requested):

src/components/molecules/Visualizer/Plugin/types.ts
keiya01: I fixed in [refactor: define Clock type in Plugin/types](https://github.com/reearth/reearth-web/pull/285/commits/7675054cdf2aeb16fcb794fd50b3fc565796c201)

src/components/molecules/Visualizer/Engine/ref.ts
keiya01: I fixed in [refactor: rename from shouldAnimate to isPlaying](https://github.com/reearth/reearth-web/pull/285/commits/89c396f37e8b076abf7bb3ab24b578ab612c48e6)

src/components/molecules/Visualizer/Engine/ref.ts
keiya01: This was fixed in [fix: use reearth event](https://github.com/reearth/reearth-web/pull/285/commits/8375e5eed05e6e89e92840f3cbd3096307e7fbb9)

src/components/molecules/Visualizer/Plugin/types.ts
```diff
@@ -346,3 +348,11 @@ export type CameraOptions = {
 
 /** Cesium API: available only when the plugin is a primitive */
 export type Cesium = {};
+
+export type Clock = {
+  startTime: Date;
+  stopTime: Date;
+  currentTime: Date;
+  tick: () => Date;
+  isPlaying: boolean;
```
rot1024, line 357:
```suggestion
  isPlaying: boolean;
  // Speed of time. Specifies a multiplier for the speed of time in reality. Default is 1.
  speed: number;
```

rot1024 (changes requested):
- Can you add speed field to clock? (it should modify Cesium Clock's multiplier)
- Can you check clock API works in plugin editor? (http://localhost:3000/plugin-editor)

src/components/molecules/Visualizer/Plugin/types.ts
keiya01: I added speed property
[fix: add speed property](https://github.com/reearth/reearth-web/pull/285/commits/1882c6d86e5dade313862cdd563e3ad1d6b413d0)

keiya01:
I tested clock in Plugin Editor by executing this script.

```js
reearth.on("tick", clock => {
  // Should fire tick event
  console.log(reearth.visualizer.clock.startTime);
  console.log(reearth.visualizer.clock.stopTime);
  console.log(reearth.visualizer.clock.currentTime);
  console.log(reearth.visualizer.clock.tick());
  console.log(reearth.visualizer.clock.isPlaying);
  console.log(reearth.visualizer.clock.speed);
});

console.log("test each fields");
console.log(reearth.visualizer.clock.startTime.toString());
console.log(reearth.visualizer.clock.startTime);
console.log(reearth.visualizer.clock.stopTime);
console.log(reearth.visualizer.clock.currentTime);
console.log(reearth.visualizer.clock.tick());
console.log(reearth.visualizer.clock.isPlaying);
console.log(reearth.visualizer.clock.speed);

reearth.visualizer.clock.isPlaying = true; // should start playing
reearth.visualizer.clock.speed = 500; // should change speed
reearth.visualizer.clock.startTime = new Date("2021-12-01"); // should change start time
reearth.visualizer.clock.stopTime = new Date("2021-12-03"); // should change stop time
reearth.visualizer.clock.currentTime = new Date("2021-12-02"); // should change current time
reearth.visualizer.clock.tick();

console.log("Should be changed");
console.log(reearth.visualizer.clock.startTime);
console.log(reearth.visualizer.clock.stopTime);
console.log(reearth.visualizer.clock.currentTime);
console.log(reearth.visualizer.clock.tick());
console.log(reearth.visualizer.clock.isPlaying);
console.log(reearth.visualizer.clock.speed);
```

rot1024 (approved):

PR #287, opened 2022-08-02
chore: migrate to Vite, upgrade Cypress to v10

rot1024:
- Introduce Vite
- Remove webpack completely
- Upgrade Cypress to v10


index.html
```diff
@@ -2,17 +2,17 @@
...
-  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <link rel="preconnect" href="https://fonts.googleapis.com">
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
   <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;700&display=swap" rel="stylesheet">
 </head>
 
 <body>
   <div id="root"></div>
+  <script type="module" src="/src/main.tsx"></script>
```
KaWaite, line 15: What is this for?

package.json
```diff
@@ -73,119 +60,96 @@
...
+        "@types/gapi.auth2": "0.0.56",
+        "@types/gapi.client": "1.0.4",
+        "@types/gapi.client.sheets": "4.0.20201029",
+        "@types/google.picker": "0.0.39",
+        "@types/tinycolor2": "1.4.2",
+        "apollo-link-sentry": "3.1.2",
+        "apollo-upload-client": "17.0.0",
+        "array-move": "4.0.0",
+        "axios": "0.27.2",
+        "cesium": "1.95.0",
```
KaWaite, line 107: Probably a stupid question, but just in case: Will `yarn upgrade interactive --latest` still work same as before with us setting specific versions here?

KaWaite: Also, is there anything I need to do when upgrading to avoid setting `^X.XX` accidentally

.gitignore
```diff
@@ -4,6 +4,7 @@ node_modules
 *.log
 /coverage
 /build
+/dist
```
KaWaite, line 7: Since we have moved from build to dist, can we remove `/build` from here?

index.html
rot1024: Read https://vitejs.dev/guide/#index-html-and-project-root

package.json
rot1024: - Of course it should work
- I found clearing yarn.lock and doing `yarn` unexpectedly installs new version of dependencies. Also renovate recommend so https://docs.renovatebot.com/configuration-options/#rangestrategy

.gitignore
rot1024: Yes, but someone remains build directory in their local machines

KaWaite (approved):

PR #294, opened 2022-08-10
fix: iframe scripts are not loaded with order

airslice:
# Overview

We used a `forEach` to trigger the execute of script inside a plugin iframe html which will lose the loading order like it is handled in a normal browser. 

## What I've done

- Replace the code with a Promise for each script so that the scripts can load in order.
- The scripts with attribute `async` `defer` or has `type="module"` (which will be handled as defer by browser) will be loaded last.
- Updated the `onLoad` part. This should be emitted after sync scripts loaded or there's no script.

## How I tested

- Test with plateauview plugin which use cdn React & ReactDOM.

## Screenshot

Before:
image

After:
image

Btw the assets issue is not related with this.

### DIFF
--- a/src/components/atoms/Plugin/IFrame/hooks.ts
+++ b/src/components/atoms/Plugin/IFrame/hooks.ts
@@ -131,29 +131,34 @@ export default function useHook({
 
     doc.body.innerHTML = html;
 
-    // exec scripts
-    Array.from(doc.body.querySelectorAll("script"))
-      .filter(
-        script =>
-          script.getAttribute("type") !== "module" &&
-          !script.getAttribute("async") &&
-          !script.getAttribute("defer"),
-      )
-      .reduce((chain, oldScript) => {
-        return chain.then(() => runScript(oldScript));
-      }, Promise.resolve())
-      .finally(() => {
-        Array.from(doc.body.querySelectorAll("script"))
-          .filter(
-            script =>
-              script.getAttribute("type") === "module" ||
-              script.getAttribute("async") ||
-              script.getAttribute("defer"),
-          )
-          .reduce((chain, oldScript) => {
-            return chain.then(() => runScript(oldScript));
-          }, Promise.resolve());
-      });
+    // exec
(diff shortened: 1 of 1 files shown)


src/components/atoms/Plugin/IFrame/hooks.ts
```diff
@@ -132,18 +132,34 @@ export default function useHook({
...
+                script.getAttribute("async") ||
+                script.getAttribute("defer"),
+            )
+            .reduce((chain, oldScript) => {
+              return chain.then(() => runScript(oldScript));
+            }, Promise.resolve());
+        });
+    }
 
     // post pending messages
```
rot1024, line 164: Send the message after the scripts are loaded. We want to send it even if the script is not present, so we want to run it when loaded is true. You can cut this out to a separate function.

KaWaite (approved):

src/components/atoms/Plugin/IFrame/hooks.ts
airslice: Thank you! I've wrapped message sending part and loaded part into a seprate function, please help re-review it. Also i've update some on async scripts i think there's no need for load them one by one.

rot1024 (approved):

src/components/atoms/Plugin/IFrame/hooks.ts
```diff
@@ -132,29 +132,50 @@ export default function useHook({
...
+          script =>
+            script.getAttribute("type") !== "module" &&
+            !script.getAttribute("async") &&
+            !script.getAttribute("defer"),
+        )
+        .reduce((chain, oldScript) => {
+          return chain.then(() => runScript(oldScript));
+        }, Promise.resolve())
+        .finally(() => {
+          onLoaded();
```
rot1024, line 149: Sorry, put it after executing scripts

rot1024 (changes requested):

rot1024 (approved):

PR #303, opened 2022-08-26
chore: update cesium to v1.96

rot1024:
Cesium was updated to v1.96 thanks to https://github.com/nshen/vite-plugin-cesium/issues/34 . I confirmed Cesium works on the Nelity deploy preview.

Also, to fix vitest failures:

- Fix dev deps versions to Increase reproducibility of test failures
- Fix node versions (Netlify and GitHub actions now reads .nvmrc)
- Update node version to v16 in `package.json` because some deps do not support node v14
- Add polyfill to import some classes from Cesium on Vitest


KaWaite (approved):
Thanks for doing this!

PR #304, opened 2022-08-26
feat: add overlay warning to earth editor if browser width is too narrow

KaWaite:
# Overview


src/i18n/translations/ja.yml
```diff
@@ -58,6 +58,8 @@ Select: 選択
 Resource URL: リソースURL
 Account Settings: アカウント設定
 Log out: ログアウト
+Your browser is too small: ''
+Re:Earth needs at least 900px width to be used effectively. Please resize your browser window.: ''
```
HideBa, line 62:
```suggestion
Your browser is too small: ブラウザ幅が小さすぎます
Re:Earth needs at least 900px width to be used effectively. Please resize your browser window.: 横幅が900px以上になるようブラウザ幅を調節してください。
```

HideBa (commented):
Added translation

PR #308, opened 2022-09-04
feat: add Navigator

keiya01:
# Overview

I added Navigator component.
But this component's design is still not fixed so this design is mock but almost features should be taken over to new design that is designed by @lavalse .

## What I've done

- implemented Navigator component that is only ui part

## What I haven't done

- implement Navigator as builtin widget.(This will be implemented in another PR)

## How I tested

1. It should rotate compass when mousedown and mousemove on compass
2. it should rotate and focus compass when mousedown and mousemove on center of compass

## Screenshot

https://user-images.githubusercontent.com/34934510/188304550-0e446c09-b253-4afc-a766-bf7cb1c37853.mov


src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+import { useNavigator } from "./hooks";
+import { CompassAngle } from "./types";
+
+export type Props = {
+  degree: number;
+  /**
+   * Pass degree of circle as callback arguments.
+   * This event is invoked when outer circle is clicked and mouse move.
+   * If click is dismissed then this event is stopped.
```
KaWaite, line 16: Sorry what does it mean if "click is dismissed"?

src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    isMovingAngle,
+    handleOnMouseDownAngle,
+    handleOnMouseDownCompass,
+  } = useNavigator({ degree, onRotate, onOrbit });
+  const t = useT();
+  return (
+    <Container>
+      <CompassContainer>
+        <Compass ref={compassRef}>
+          <>
```
KaWaite, line 59: Do you need this fragment? 
I've seen other developers use it where I feel its not needed, so I am curious if this is just a mistake or you are using it purposefully.

src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+            <CompassIcon onMouseDown={handleOnMouseDownCompass}>
+              <Icon
+                icon="compass"
+                color="#000"
+                aria-label={t("compass")}
+                size={64}
+                style={{ transform: `rotate(${compassDegree}deg)` }}
+              />
+            </CompassIcon>
```
KaWaite, line 68: Are you able to make a `const CompassIcon = styled(Icon)`...styles`;` and avoid wrapping the Icon in a div? Or does the Icon component not give you access to onMouseDown?
If the styled(Icon) is possible, it reduces our nested div count a little.. :P

src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+            {isMovingAngle && (
+              <CompassFocusIcon
+                style={{
+                  transform: `rotate(${compassFocusAngle.degree}deg)`,
+                }}
+                data-testId="compassFocus">
+                <Icon icon="compassFocus" color="blue" alt="" size={30} />
+              </CompassFocusIcon>
```
KaWaite, line 76: Same as the comment above. You can resolve this comment if it seems appropriate.

src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+            )}
+            <AngleIcon onMouseDown={handleOnMouseDownAngle}>
+              <Icon icon="navigatorAngle" aria-label={t("adjust angle")} size={32} />
+            </AngleIcon>
```
KaWaite, line 80: Same here

src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+          </>
+        </Compass>
+        <Help onClick={onClickHelp}>?</Help>
+      </CompassContainer>
+      <Tool>
+        <>
```
KaWaite, line 86: Same as the above fragment comment.

src/components/atoms/Navigator/index.tsx
KaWaite: I noticed you are doing this wrapper with all the Icons in this component. If you can do styled(Icon), please update all the other icons too, which will reduce the component size by quite a bit

src/components/atoms/Navigator/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  return (
+    <Container>
+      <CompassContainer>
+        <Compass ref={compassRef}>
+          <>
+            <CompassIcon onMouseDown={handleOnMouseDownCompass}>
+              <Icon
+                icon="compass"
+                color="#000"
+                aria-label={t("compass")}
```
KaWaite, line 64: Since aria-label's are unique and seem to not need grammar, in-depth descriptions, etc, can we consider doing `aria-label-someNameOrDescription`?
That way it'll be a bit clearer in the translation files. It DOES mean we have to update with the ACTUAL desired value inside the translation file. So currently, we just copy and paste the key to the value in the English.
Please feel free to disagree with this approach or if you have a better idea please let me know!

src/components/atoms/Navigator/index.tsx
keiya01: There are some reason why I don't use `styled(Icon)` pattern like the following:pray:

- `<AngleIcon/>` and `<CompassIcon/>` need to set `onMouseDown` event, but `<Icon/>` does not take event.
- `<CompassFocusIcon/>` has different size to `<Icon />` size. And `<CompassFocusIcon/>` has animation that rotate around the center of compass while `<CompassFocusIcon/>` keep size. To accomplish this animation, we need to define `<CompassFocusIcon/>`.
- `<ToolIconButton/>` is button element.

src/components/atoms/Navigator/index.tsx
keiya01: Does what you mean that we should use the special key for `aria-label` like the following?

use special key

```js
aria-label={t("aria-label-compass")}
```

and set to translation file

```yaml
"aria-label-compass": compass
```

src/components/atoms/Navigator/index.tsx
keiya01: I fixed!
[refactor: remove fragment](https://github.com/reearth/reearth-web/pull/308/commits/8e645e54eef3f9d5fa4269b34f9219cea30de9a0)

src/components/atoms/Navigator/index.tsx
keiya01: This was fixed with the above comment.

src/components/atoms/Navigator/index.tsx
keiya01: Sorry, I fixed this comment...
[chore: fix props description](https://github.com/reearth/reearth-web/pull/308/commits/e547b5673e3bd07cb81c0b2ece3697454c491d1e)

src/components/atoms/Navigator/index.tsx
KaWaite: @keiya01 Thanks for being careful, but if you have the time I think it would be great to update the Icon component to accept onMouseDown/Up events. 
For the rotation, can you not just set transform in the styled icon: `styled(Icon)`HERE``? If not, maybe adding a transition prop to Icon would work?
If you find an atom component doesn't meet your needs, and adding a prop seems beneficial to the atom AND doesn't over complicate it feel free to modify them when you need!

src/components/atoms/Navigator/index.tsx
KaWaite: Yes! Exactly!

src/components/atoms/Navigator/index.tsx
keiya01: https://github.com/reearth/reearth-web/pull/308/commits/cd21558514e20b788c00f28fa295599f2edfb530
I fixed this review!

src/components/atoms/Navigator/index.tsx
keiya01: https://github.com/reearth/reearth-web/pull/308/commits/ad87c866ed63aa40a82916eb8816e286eb946c52
I fixed this!

src/components/atoms/Navigator/index.tsx
keiya01: Hmm, if I set `onMouseDown` event to svg element then test fail... But `Navigator` works fine in storybook.
I don't know why this behavior occurs this problem but maybe this is occurred by jest-dom.
Do you know about this behavior?

keiya01: Testing svg element is difficult so I think we can keep `div` element. Do you think about this?

KaWaite: I have noticed another unexpected behaviour around Icons when passing some props to an Icon using styled components, so I am thinking of removing the svg dep we are currently using and try making the Icons component to be similar to what we are doing in another repo, that seems to be working well.
For you and this PR, if it works in storybook well, and nothing breaks if you extensively use it there, then I think it is fine!

KaWaite (approved):

PR #310, opened 2022-09-06
feat: extend plugin api with captureScreen

airslice:
# Overview

This feature enables plugin get the image of the map.

Without this API the plugin can manually get the canvas and use `canvas.toDataURL` but the result will be a full black image due to the canvas webgl settings.

The solutions are:
A: Config contextOptions with `{{ webgl: { preserveDrawingBuffer: true } }}` but this will technically harm the performance so i did not apply this.
B: Call `canvas.toDataURL` immediately after a `viewer.render`. This is what i implemented for the API.

**NOTICE**: the output image comes from the canvas so anything implemented out of canvas (eg. infobox) can not be included. If you want to take a screenshot of all HTML elements you may try implement with some third-party libraries in plugin code.

## What I've done

Extend the plugin API with 
```
reearth.scene.captureScreen: (type?: string, encoderOptions?: number) => string | undefined
```
Refer `canvas.toDataURL` for parameters `type` `encoderOptions`

Thers's also a little refactor to unit test file included in this PR (Basically renaming function).

## How I tested

- Add unit test.
- Test with console.
- Test with plugin. Here's an example which can be tested with plugin editor.
```
reearth.ui.show(`
  <div id="wrapper">
    <button id="button">Capture</button>
  </div>
  <script>
    document.getElementById("button").addEventListener("click", (e) => {
      parent.postMessage({ type: "captureScreen" }, "*");
    });
    addEventListener("message", e => {
      if (e.source !== parent) return;
      if (e.data.type && e.data.type === 'getCaptureScreen') {
        const fileName = "capture.png";
        const link = document.createElement("a");
        link.download = fileName;
        link.href = e.data.payload;
        link.click();
        link.remove();
      }
    })
  </script>
`);

reearth.on("message", msg => {
  if (msg.type === "captureScreen") {
    reearth.ui.postMessage({
      type: "getCaptureScreen",
      payload: reearth.scene.captureScreen(),
    });
  }
});
```

### DIFF
--- a/src/components/molecules/Visualizer/Engine/Cesium/useEngineRef.test.tsx
+++ b/src/components/molecules/Visualizer/Engine/Cesium/useEngineRef.test.tsx
@@ -9,9 +9,6 @@ import { EngineRef } from "../ref";
 
 import useEngineRef from "./useEngineRef";
 
-const fn = vi.fn(e => e);
-const props = { x: 1, y: 1 };
-
 test("engine should be cesium", () => {
   const { result } = renderHook(() => {
     const cesium = useRef<CesiumComponentRef<CesiumViewer>>(null);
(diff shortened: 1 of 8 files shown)


lavalse:
I have quickly tested. Seems working fine.

rot1024 (approved):

KaWaite (approved):

PR #316, opened 2022-09-13
fix: timeline width exceeding browser width

KaWaite:
# Overview


src/components/atoms/Timeline/ScaleList.tsx
```diff
@@ -12,7 +12,7 @@ type Props = {
 
 const ScaleList: React.FC<Props> = ({ gapHorizontal, ...props }) => {
   return (
-    <ScaleContainer style={{ gap: `0 ${gapHorizontal}px` }}>
+    <ScaleContainer gapHorizontal={gapHorizontal}>
```
keiya01, line 15: I think this should keep using style because `gapHorizontal` is often updated. And if style is used then test should be passed.

PR #321, opened 2022-09-21
fix: remove `show` props from modal extension

HideBa:
## What I've done
Removed `show` props from modal extension bc it's not needed.


KaWaite (approved):
Thanks for cleaning up a bit ;)

PR #324, opened 2022-09-30
fix: Camera popup icon is too small and the icon on storytelling

nina992:
# Overview
The camera icon is too small as below 
image
also the icon in widget story telling is also too small 
image

## What I've done
resize icons 

## Screenshot
size =24 
image

-----------
size = 28 
image

## Memo
https://github.com/reearth/reearth/issues/344


nina992:
@lavalse  would please check sizes

lavalse:
image
The icon size 24 x 24 is too large. Please keep in 16 x 16 px 🙏

lavalse:
and also not sure for the camera popup size. For if you can also keep them as 16 x 16px will be quite nice. 🙏

src/components/molecules/EarthEditor/PropertyPane/PropertyField/CameraField/index.tsx
```diff
@@ -262,10 +262,13 @@ const Popup = styled.ul<{ open: boolean }>`
...
 const FormGroup = styled.div`
   display: flex;
+  align-item: center;
 `;
 
 const FormIcon = styled(Icon)`
-  margin: 10px 10px 10px 0;
+  margin-top: auto;
+  margin-bottom: 25px;
+  width: 14.07%;
```
KaWaite, line 271: If there is no support for resizing/mobile, etc please use pixels instead of percentage! 
But, in addition since you are modifying an Icon in this case, please use the Icon's size prop (as a pixel value)

nina992: as dicused before the width property doesn't work right in this case unless I change the width of the popup itself , removing marging or padding affects only few pixels and I don't wan't to make any changes on icon componenent itself , so I suggest anther solution would please check , also I replace percentage with pixels

KaWaite (approved):

PR #332, opened 2022-10-19
fix: zoom to layer functionality

nina992:
# Overview
When pressing zoom to layer and then scrolling out or moving the camera out of the layer , then trying to zoom again , the zoom button doesn't work.
## What I've done
adding state to fix that


src/components/molecules/Visualizer/hooks.ts
```diff
@@ -75,6 +77,7 @@ export default ({
 }) => {
   const engineRef = useRef<EngineRef>(null);
   const [overriddenSceneProperty, overrideSceneProperty] = useOverriddenProperty(sceneProperty);
+  const [, zoomToLayer] = useZoomedLayerId();
```
KaWaite, line 80: So somewhere else we are doing `const [zoomedLayerId] = useZoomedLayerId();`, and then here you are doing `const [, zoomToLayer] = useZoomedLayerId();`
I feel like it makes more sense to declare them in the same place (not sure here or where `zoomedLayerId` is currently declared is better. Please look at the code and decide which is better)

src/components/molecules/Visualizer/hooks.ts
```diff
@@ -3,6 +3,8 @@ import { useRef, useEffect, useMemo, useState, useCallback, RefObject, useReduce
 import { initialize, pageview } from "react-ga";
 import { useSet } from "react-use";
 
+// eslint-disable-next-line no-restricted-imports
```
KaWaite, line 6: Our import restrictions are pretty specific and deliberate. Please don't ignore them.

src/components/molecules/Visualizer/hooks.ts
KaWaite: Please see above comment about import restrictions.

src/components/molecules/Visualizer/hooks.ts
nina992: I noticed that they ignord ather place ,so I continued with that

KaWaite (approved):

PR #338, opened 2022-10-27
feat: add clipping box

keiya01:
# Overview

I added clipping box feature.
This clip the 3dtileset inside the box.

[Clipping box plugin is here.](https://github.com/eukarya-inc/reearth-plugin-clipping-box)

## What I've done

- Add clipping feature in Tileset component
- Add the Box component
- Add lookAt without animation

## What I haven't done

- The feature that scale the Box

## How I tested

1. Install the [clipping-box-plugin](https://github.com/eukarya-inc/reearth-plugin-clipping-box)
2. And set position, tileset url, box color in the widget panel.
3. Move the Box horizontally or vertically.

## Screenshot

https://user-images.githubusercontent.com/34934510/198632757-c38d2484-7fcd-434e-9f74-a90ef02a2c18.mov


src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+};
+
+const Box: React.FC<PrimitiveProps<Property>> = memo(function BoxPresenter({ layer }) {
+  const { id, isVisible, property } = layer ?? {};
+  const { fillColor, outlineColor, outlineWidth, fill, outline } = property?.default ?? {};
+
+  const [trs] = useState(() => updateTrs(new TranslationRotationScale(), property));
+  useEffect(() => {
+    updateTrs(trs, property);
+  }, [property?.default?.position, property?.default?.location, property?.default?.height, property?.default?.dimensions, property?.default?.dimensions?.y, property?.default?.dimensions?.z]);
```
keiya01, line 144: This is for performance.
Updating the state too many times will result in poor performance, so update the reference of `trs`.

src/components/molecules/Visualizer/Engine/Cesium/Tileset/index.tsx
```diff
@@ -13,25 +26,109 @@ export type Property = {
...
+        hpr
+          ? Matrix4.fromRotationTranslation(Matrix3.fromHeadingPitchRoll(hpr), position)
+          : Transforms.eastNorthUpToFixedFrame(position),
+        Matrix4.fromScale(dimensions, new Matrix4()),
+        new Matrix4(),
+      );
+
+      const inverseOriginalModelMatrix = Matrix4.inverse(clippingPlanesOriginMatrix, new Matrix4());
+
+      Matrix4.multiply(inverseOriginalModelMatrix, boxTransform, clippingPlanes.modelMatrix);
```
keiya01, line 126: This is also the same with the above comment.
The reference for `clippingPlanes.modelMatrix` get updated.

src/components/molecules/Visualizer/Engine/Cesium/common.ts
```diff
@@ -278,6 +279,12 @@ export const lookAt = (
   };
 };
 
+export const lookAtWithoutAnime = (scene: Scene) => {
```
KaWaite, line 282: I kinda love this name haha. But I think, though it will be really long, should be `lookAtWithoutAnimation`.
`Anime` for English speakers is just Japanese animation :P

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  Cartesian3.clone(
+    new Cartesian3(dimensions?.x || 100, dimensions?.y || 100, dimensions?.z || 100),
+    trs.scale,
+  );
+
+  return trs;
+};
+
+const Box: React.FC<PrimitiveProps<Property>> = memo(function BoxPresenter({ layer }) {
```
KaWaite, line 137: <PrimitiveProps<Property>> can just be `Props`, no? (above its declared like that)

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        <Side
+          key={`${id}-plane-${SIDE_PLANE_NAMES[i]}`}
+          id={`${id}-plane-${SIDE_PLANE_NAMES[i]}`}
+          planeLocal={plane}
+          style={style}
+          trs={trs}
+        />
+      ))}
+    </>
+  );
```
KaWaite, line 169: Unless there is a specific need for the fragments, you can write it like this:

```
  return !isVisible ?  SIDE_PLANES.map((plane, i) => (
        <Side
          key={`${id}-plane-${SIDE_PLANE_NAMES[i]}`}
          id={`${id}-plane-${SIDE_PLANE_NAMES[i]}`}
          planeLocal={plane}
          style={style}
          trs={trs}
        />
      )) : null
```

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+  id: string;
+  planeLocal: CesiumPlane;
+  trs: TranslationRotationScale;
+  style: BoxStyle<Color>;
+}> = memo(
+  function SidePresenter({ id, planeLocal, style, trs }) {
+    const normalAxis = planeLocal.normal.x ? Axis.X : planeLocal.normal.y ? Axis.Y : Axis.Z;
+    const cbRef = useMemo(
+      () => new CallbackProperty(() => trs.translation, false) as unknown as PositionProperty,
+      [trs],
```
KaWaite, line 71: Is there any way to type this better? Or can you explain why we must typecast as unknown before typing as PositionProperty?

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,172 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
KaWaite, line 1: Can you use `// eslint-disable-line react-hooks/exhaustive-deps` only where it is necessary instead of on the whole file? I feel like ignoring on the whole file is kinda dangerous

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+  } else if (planeNormalAxis === Axis.Y) {
+    planeDimensions.x = boxDimensions.x;
+    planeDimensions.y = boxDimensions.z;
+  } else if (planeNormalAxis === Axis.Z) {
+    planeDimensions.x = boxDimensions.x;
+    planeDimensions.y = boxDimensions.y;
+  }
+}
+
+const Side: FC<{
```
KaWaite, line 61: Can you split Side in to another file and then import it in here? There is a lot of logic between Side and Box and it would be nice to keep things a little more segregated.

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: I think we should use PrimitiveProps.
PrimitiveProps indicates an interface which a reearth layer can take.

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: The component need to take `ReactElement` type, but TypeScript infer that not wrapped array of components with fragment is `Element[]` type, so we need to wrap array of element with fragment.

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: This is for performance. If we don't use `CallbackProperty`, we need to reconstruct the trs, and PlaneGraphic component need rerender, but this is slow as I commented in [here](https://github.com/reearth/reearth-web/pull/338#discussion_r1008951750).

But CallbackProperty is not officially supported as [discussed in this issue](https://github.com/CesiumGS/cesium/issues/8944), so we need to use workaround for TypeScript.
Using `CallbackProperty` for entity's position is not officially supported, but [terriajs is using this workaround in here](https://github.com/TerriaJS/terriajs/blob/cad62a45cbee98c7561625458bec3a48510f6cbc/lib/Models/BoxDrawing.ts#L708), so I think this is no problem.

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: I fixed in [chore: remove unnecessary disabling eslint comment](https://github.com/reearth/reearth-web/pull/338/commits/6f65b3b54d9e32cd1038fcdd466558aecc39c991) :pray:

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: I fixed in [refactor: split Side to other file](https://github.com/reearth/reearth-web/pull/338/commits/8305e3060d4ceee9b701b2029b32085b127baef9)

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
KaWaite: Fair enough. If that is the case, is the exported Props type (on line 10) used in another component? If not, please remove the export. I see you import `Props` from `../../Primitive` in this component and do the same thing in `Tileset/index` so I am guessing it might not be necessary

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
KaWaite: Ok, gotcha!

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
KaWaite: 頭いいよね :P
Ok!

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+    position?: LatLng;
+    location?: LatLng;
+    height?: number;
+    dimensions?: {
+      x: number;
+      y: number;
+      z: number;
+    };
+  } & BoxStyle<string>;
+};
```
rot1024, line 31:
```suggestion
export type Property = {
  default?: {
    location?: LatLngHeight;
    width?: number;
    height?: number;
    length?: number;
    heading?: number;
    pitch?: number;
    roll?: number;
    fillColor?: string;
    outlineColor?: string;
    outlineWidth?: number;
    fill?: boolean;
    outline?: boolean;
  };
};
```

src/components/molecules/Visualizer/Engine/Cesium/Tileset/index.tsx
```diff
@@ -13,25 +26,109 @@ export type Property = {
     tileset?: string;
     styleUrl?: string;
     shadows?: "disabled" | "enabled" | "cast_only" | "receive_only";
+    edgeWidth?: number;
+    edgeColor?: string;
+    clippingPlaneCollection?: ClippingPlaneCollection;
```
rot1024, line 31:
```suggestion
    EXPERIMENTAL_clipping?: EXPERIMENTAL_Clipping;
```

src/util/value.ts
```diff
@@ -46,6 +46,37 @@ export type Rect = {
...
+    /**
+     * z-axis
+     */
+    height?: number;
+  };
+  heading?: number;
+  pitch?: number;
+  roll?: number;
+  height?: number;
+} & LatLng;
```
rot1024, line 78:
```suggestion
// ideal for plugin developers, but it's hard to implement it with Cesium
export type Plane = {
  location: LatLngHeight;
  width: number;
  height: number;
  heading: number;
  pitch: number;
};

// Familiar with Cesium
export type EXPERIMENTAL_Clipping = {
  planes: { normal: [number, number, number]; distance: number }[];
  modelMatrix: number[];
}
```

rot1024 (changes requested):

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: Yes, it is. The props is used in storybook for the box.

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: This was fixed in [refactor: fix property type](https://github.com/reearth/reearth-web/pull/338/commits/9cecd3a70245748ab056c2d9ce0a54dda57be7b0)

src/components/molecules/Visualizer/Engine/Cesium/Tileset/index.tsx
keiya01: This was fixed in [fix: clipping collection to experimental](https://github.com/reearth/reearth-web/pull/338/commits/bf8000046bbbfb2d545d64a68bd4a8c2b9e7e2a9)

src/util/value.ts
keiya01: I fixed this comment. [fix: use experimental clipping](https://github.com/reearth/reearth-web/pull/338/commits/60d8d4b961f99da27b7f165688166bd729c2b7a2)

KaWaite (approved):

rot1024 (approved):

PR #340, opened 2022-10-28
chore: fix storybook build was failed with node 18

rot1024:
fix storybook build was failed with node 18


rot1024:
We will have to wait for the release of storybook v7 which does not use webpack4 to fundamentally solve the problem.

PR #343, opened 2022-10-31
feat: support acquiring locations with terrain

airslice:
# Overview

Mouse events want to get the correct location when terrain is on.

## What I've done

Update the common `getLocationFromScreenXY` function, added a prop `withTerrain` default as `false`.

If `withTerrain` is `true` it will use `globe.pick` instead of `camera.pickEllipsoid` so that terrain can take effect. And if doesn't get the location it will still try `camera.pickEllipsoid` as a polyfill.

## How I tested

Tested with plugin using mouse event.

## Screenshot

image

### DIFF
--- a/src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
+++ b/src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
@@ -213,7 +213,9 @@ export default ({
         const props: MouseEvent = {
           x: position?.x,
           y: position?.y,
-          ...(position ? getLocationFromScreenXY(viewer.scene, position.x, position.y) ?? {} : {}),
+          ...(position
+            ? getLocationFromScreenXY(viewer.scene, position.x, position.y, true) ?? {}
+            : {}),
         };
         const layerId = getLayerId(target);
         if (layerId) props.layerId = layerId;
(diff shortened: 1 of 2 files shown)


rot1024 (approved):

src/components/molecules/Visualizer/Engine/Cesium/hooks.ts
```diff
@@ -213,7 +213,9 @@ export default ({
         const props: MouseEvent = {
           x: position?.x,
           y: position?.y,
-          ...(position ? getLocationFromScreenXY(viewer.scene, position.x, position.y) ?? {} : {}),
+          ...(position
+            ? getLocationFromScreenXY(viewer.scene, position.x, position.y, true) ?? {}
```
KaWaite, line 217: Why are you setting withTerrain to true here on all mouse events with position? If you hardcode that, is there a reason to having the withTerrain prop?

rot1024: This is the difference between the interest that the plug-in has and the interest that the map engine has. Plug-ins only get the basic terrain-enabled location, but there may be times when you want to do pure location acquisition inside Re:Earth, independent of the plug-in. In that case, you are still calling a function exposed by the map engine. There is no reason to go to the trouble of making changes that narrow the degree of freedom.

KaWaite (approved):

PR #344, opened 2022-10-31
fix: alignment issues in Widget Align System

KaWaite:
# Overview
Placement of widgets were incorrect in a couple grid areas. 
Needed to upgrade react-align since there were issues in that package. 
Also updated on this end some logic errors.


airslice (approved):
The align logic seems a little complex and i think i only kind of understand it. I asked red to help test some and everything seems working good!

PR #348, opened 2022-11-04
feat: extend plugin API supports getting pixel distance

airslice:
# Overview

Plugin need to get the pixel distance to implement a distance legend.

This API will get two points on earth surface base on ray from two adjacent pixel on bottom (as default) | top | center of the screen and return the distance in meters.

## What I've done

Implemnt API:
```ts
reearth.camera.getPixelDistance(verticalPosition?: "top" | "middle" | "bottom" ) => number | undefined
```

`verticalPosition` defines the y position of sample point.
- `top`: top center of the screen space.
- `middle`: middle center of the screen space.
- `bottom` (as default): bottom center of the screen space.

## How I tested

Test with devtool console.
Test with plugin editor using:

```js
reearth.ui.show(`
  <style>
    html,
    body {
      margin: 0;
      width: 350px;
      font-size: 14px;
    }
    button {
      margin: 0;
    }
    #wrapper {
      height: 100%;
      box-sizing: border-box;
      border-radius: 12px;
      padding: 12px;
      background: white;
    }
    #wrapper {
      box-sizing: border-box;
      border-radius: 12px;
      background: white;
      height: 100%;
    }
    h3{
      margin: 0 0 5px;
    }
    h4{
      margin: 5px 0;
    }
    #currentUnitLine{
      height: 2px;
      background: #000;
      width: 0;
    }
  </style>

  <div id="wrapper">
    <h3>Distance Legend</h3>
    <h4 id="pixelDistance"></h4>
    <h4 id="currentLabel"></h4>
    <div id="currentUnitLine"></div>
  </div>

  <script>
    const distances = [
      1, 2, 3, 5, 10, 20, 30, 50, 100, 200, 300, 500, 1000, 2000, 3000, 5000, 10000,
      20000, 30000, 50000, 100000, 200000, 300000, 500000, 1000000, 2000000,
      3000000, 5000000, 10000000, 20000000, 30000000, 50000000
    ];

    window.addEventListener("message", function (e) {
      if (e.source !== parent) return;
      document.getElementById("pixelDistance").innerHTML = e.data.payload;

      const pixelDistance = e.data.payload ?? 0;
      
      const maxBarWidth = 100;
      let distance;
      for (let i = distances.length - 1; !distance && i >= 0; --i) {
        if (distances[i] / pixelDistance < maxBarWidth) {
          distance = distances[i];
        }
      }

      let label;
      if (distance >= 1000) {
        label = (distance / 1000).toString() + " km";
      } else {
        label = distance.toString() + " m";
      }
      document.getElementById("currentLabel").innerHTML = label;
      document.getElementById("currentUnitLine").style.width = (distance / pixelDistance) + 'px';
      
    });
  </script>
`);

reearth.on("cameramove", () => {
  reearth.ui.postMessage({
    type: 'getPixelDistance',
    payload: reearth.camera.getPixelDistance()
  }, "*");
});
```

Please confirm if it's okey to set this API as `reearth.camera.getPixelDistance`


src/components/molecules/Visualizer/Engine/Cesium/common.ts
```diff
@@ -633,6 +634,32 @@ export async function flyToGround(
   );
 }
 
+export function getPixelDistance(viewer: Viewer, sample = "bottom") {
+  const scene = viewer.scene;
+  const camera = scene?.camera;
```
KaWaite, line 639: Currently you aren't using this camera variable (you probably aren't getting a warning because you do the if !camera check).
You are also using canvas and globe from the scene below, so you can clean up by doing

`const { camera, canvas, globe } = scene;`

Whether you do this suggestion or not, please use camera where you are declaring `left` and `right` ;)

src/components/molecules/Visualizer/Plugin/types.ts
```diff
@@ -362,8 +362,11 @@ export type Camera = {
     options?: CameraOptions,
     offset?: number,
   ) => void;
+  readonly getPixelDistance: (sample?: PixelDistanceSamplePosition) => number | undefined;
```
KaWaite, line 365: Is it really a `sample`? It is actually the y-location, or verticalPosition, right? Can you think of a better name for the param?

src/components/molecules/Visualizer/Engine/Cesium/common.ts
airslice: Updated. I found some check or definition is unnessary and removed them.

src/components/molecules/Visualizer/Plugin/types.ts
airslice: Updated to `verticalPosition`. I think the full name might be `verticalSamplePosition` and the type name might be `PixelDistanceVerticalSamplePosition` but this is really too long and `PixelDistanceSamplePosition` seems more meaningful than `PixelDistanceVerticalPosition` so i did not change the type name.

src/components/molecules/Visualizer/Engine/Cesium/common.ts
```diff
@@ -633,6 +634,29 @@ export async function flyToGround(
   );
 }
 
+export function getPixelDistance(scene: Scene, verticalPosition = "bottom") {
+  const width = scene.canvas.clientWidth;
+  const height = scene.canvas.clientHeight;
+
+  const sampleHeight =
```
KaWaite, line 641: I checked briefly cesium docs and the getPickRay's second param is where we are using `sampleHeight` is `result`, which it explains is the object that we store the result. Which makes me alittle confused why we are passing a number in to there.
Can you explain that a little more? And also think of a better name than sampleHeight?

src/components/molecules/Visualizer/Plugin/types.ts
KaWaite: Is there any reason you are wanting to keep Sample in the type name? For me I can't see why its a sample. (kind of related to my other comment, so you can reply to both there)

src/components/molecules/Visualizer/Engine/Cesium/common.ts
airslice: Actually `sampleHeight` is the second param of `new Cartesian2` not `getPickRay`. The reason why i want to use the word `sample` is that in most cases our camera is a perspective camera, which will have different pixel distance in different point and we need to pick one point as a sample. Here the `sampleHeight` actually means the sample point's height (or top better?) on screen space. Emmm maybe need to change widht/height to left/top?

src/components/molecules/Visualizer/Plugin/types.ts
airslice: Please check the reply below.

airslice:
This API seems too specific to a specific use case. We decide to cancel this and instead expose `getLocationFromScreenPosition`. Please check: https://github.com/reearth/reearth-web/pull/350

PR #351, opened 2022-11-09
feat: extend plugin API supports get scene inEditor

airslice:
# Overview

Plugin needs to know whether it's running with the earth editor.

For pages:
`earth editor` => `true`
`preview` => `false`
`published` => `false`
`plugin editor` => `false` for now. We can also update plugin editor's UI to support switch value if needed in the future.

## What I've done

Implement plugin API

```ts
reearth.scene.inEditor: boolean
```


src/components/molecules/Visualizer/hooks.ts
```diff
@@ -251,6 +253,7 @@ export default ({
     {
       engineName: engineType || "",
       sceneProperty: overriddenSceneProperty,
+      isInEditor: isInEditor ?? false,
```
KaWaite, line 256:
```suggestion
      isInEditor: !!isInEditor,
```

src/components/organisms/EarthEditor/CanvasArea/index.tsx
```diff
@@ -61,6 +62,7 @@ const CanvasArea: React.FC<Props> = ({ className, isBuilt }) => {
       engine="cesium"
       isEditable={!isBuilt}
       isBuilt={!!isBuilt}
+      isInEditor={isInEditor !== false}
```
KaWaite, line 65:
```suggestion
      isInEditor={!!isInEditor}
```

src/components/pages/Preview/index.tsx
```diff
@@ -24,7 +24,7 @@ const PreviewPage: React.FC<Props> = () => {
     <ThemeProvider>
       <DndProvider>
         <AuthenticationRequiredPage>
-          <CanvasArea isBuilt />
+          <CanvasArea isBuilt isInEditor={false} />
```
KaWaite, line 27: Just don't pass isInEditor and by handling it in the CanvasArea like `!!isInEditor` it'll be false

src/components/molecules/Visualizer/Plugin/api.ts
```diff
@@ -260,6 +261,7 @@ export function commonReearth({
   cameraViewport: () => GlobalThis["reearth"]["visualizer"]["camera"]["viewport"];
   captureScreen: GlobalThis["reearth"]["scene"]["captureScreen"];
   getLocationFromScreen: GlobalThis["reearth"]["scene"]["getLocationFromScreen"];
+  isInEditor: () => GlobalThis["reearth"]["scene"]["inEditor"];
```
rot1024, line 264:
```suggestion
  InEditor: () => GlobalThis["reearth"]["scene"]["inEditor"];
```

rot1024 (changes requested):
Unify isInEditor to inEditor

src/components/organisms/EarthEditor/CanvasArea/index.tsx
airslice: Here i want to set the value to `true` except got `isInEditor` as `false`; `CanvasArea` is a child component of `EarthEditor` so it should always be `true` for `isInEditor`, but i found the implementation of preview page as it use the `CanvasArea` directly therefore i pass the params `false` from preview page. So that's why i did not use `!!isInEditor`

src/components/molecules/Visualizer/hooks.ts
airslice: Updated.

src/components/organisms/EarthEditor/CanvasArea/index.tsx
airslice: Updated.

src/components/pages/Preview/index.tsx
airslice: Updated. And also set `<CanvasArea inEditor />` in `page/EarthEditor`.

src/components/molecules/Visualizer/Plugin/api.ts
airslice: Updated.

rot1024 (approved):

KaWaite (approved):

PR #357, opened 2022-11-15
feat: editable box

keiya01:
# Overview

I added some layer to the Box layer. These layer is used to handle some interaction like the following.
- Scaling the box
- Rotating the box

## What I've done

- added `ScalePoint` layer
- added `Edge` layer
- added `boxscale` event to handle scaling box
- added `boxrotate` event to handle rotating box

## How I tested

1. Install the [clipping-box-plugin](https://github.com/eukarya-inc/reearth-plugin-clipping-box)
2. And set position, tileset url, box color in the widget panel.
3. Move the Box horizontally or vertically.

## Screenshot

https://user-images.githubusercontent.com/34934510/201908204-3ca53edf-7825-455e-b4c3-e619dbc61657.mp4


src/components/molecules/Visualizer/Plugin/context.tsx
```diff
@@ -82,10 +82,16 @@ export type Props = {
   flyToGround: (destination: FlyToDestination, options?: CameraOptions, offset?: number) => void;
 };
 
+type SelectedRearthEventType = Pick<
```
KaWaite, line 85:
```suggestion
type SelectedReearthEventType = Pick<
```

src/components/molecules/Visualizer/Engine/Cesium/Box/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+  Quaternion,
+  Ray,
+  Scene,
+  Transforms,
+  TranslationRotationScale,
+} from "cesium";
+
+import { Property } from ".";
+
+const translationWithClamping = (
```
KaWaite, line 18: You have this in the Tileset file, too. Since this declaration is in a utils file, can you just export this one and use it in Tileset too?
It might make sense to move it in to the Cesium utils to be accessed by both

src/components/molecules/Visualizer/Engine/Cesium/Box/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  const farPoint2d = scene.cartesianToCanvasCoordinates(
+    Ray.getPoint(ray, length),
+    new Cartesian2(),
+  );
+  const screenVector2d = Cartesian2.subtract(farPoint2d, nearPoint2d, result);
+  return screenVector2d;
+}
+
+const dotProductMousePosition = (
```
KaWaite, line 86: First, I feel like `dotMousePosition` is maybe descriptive enough, no?
Secondly, this returns pixels per step as well as move amount pixels. Would you consider this information together is a position? Or is it data that can be used to get the position? Does that make sense?

src/components/molecules/Visualizer/Engine/Cesium/Box/utils.ts
```diff
@@ -0,0 +1,152 @@
...
+  const pixelsPerStep = Cartesian2.magnitude(screenVector2d);
+  const moveAmountPixels = Cartesian2.dot(mouseVector2d, screenNormal2d);
+
+  return {
+    pixelsPerStep,
+    moveAmountPixels,
+  };
+};
+
+export const computeMoveAmount = (
```
KaWaite, line 114: It'll be kinda long, but should this be `computeMouseMoveAmount`? Or `computeMouseMove`?

src/components/molecules/Visualizer/Engine/Cesium/Box/Edge.tsx
```diff
@@ -0,0 +1,129 @@
...
+    () =>
+      new CallbackProperty(() => {
+        const position1 = new Cartesian3();
+        const position2 = new Cartesian3();
+        const matrix = Matrix4.fromTranslationRotationScale(trs);
+        Matrix4.multiplyByPoint(matrix, edge.start, position1);
+        Matrix4.multiplyByPoint(matrix, edge.end, position2);
+        return [position1, position2];
+      }, false),
+  );
```
KaWaite, line 68: Sorry maybe a really stupid question/request, but just want to ask about this pattern (never seen useState used in this way before). Does this basically run the callback inside the useState() on render and assign it to the value (in this case, cbp)??
If so, why are you using useState like this, and not using useMemo?

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -20,75 +35,342 @@ export type Property = {
...
 
-  Cartesian3.clone(new Cartesian3(width || 100, length || 100, height || 100), trs.scale);
+  const [terrainHeightEstimate, setTerrainHeightEstimate] = useState(0);
+  const [trs] = useState(() =>
+    updateTrs(new TranslationRotationScale(), property, terrainHeightEstimate),
+  );
 
-  return trs;
-};
+  const isProgressSamplingTerrainHeight = useRef(false);
```
KaWaite, line 94: Is this suppose to be `inProgressSamplingTerrainHeight`? This is just keeping track of whether we are calculating the height, right?

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
```diff
@@ -20,75 +35,342 @@ export type Property = {
...
-const Box: React.FC<PrimitiveProps<Property>> = memo(function BoxPresenter({ layer }) {
-  const { id, isVisible, property } = layer ?? {};
-  const { fillColor, outlineColor, outlineWidth, fill, outline } = property?.default ?? {};
+    isProgressSamplingTerrainHeight.current = true;
+    if (keepBoxAboveGround) {
+      sampleTerrainHeight(viewer.scene, trs.translation).then(v => {
+        setTerrainHeightEstimate(v ?? 0);
+        isProgressSamplingTerrainHeight.current = false;
+      });
+    }
```
KaWaite, line 106: Specific to this callback: If we call this callback and `keepBoxAboveGround` is set to false, we won't calculate a new sampleTerrainHeight, and `isProgressSamplingTerrainHeight` will stay true, which means when we run this function again, it'll always return at the top and NEVER become false again, right?

Specific to this component: We use this callback in the useEffect, where we run this component IF `keepBoxAboveGround`. If we only use this function when `keepBoxAboveGround` is true, you can remove this if statement inside the function, right? That'll solve the above problem I mentioned.

src/util/value.ts
```diff
@@ -82,6 +82,7 @@ export type EXPERIMENTAL_clipping = {
   heading?: number;
   pitch?: number;
   roll?: number;
+  keepAboveGround?: boolean;
```
KaWaite, line 85: Our system's default is more or less don't allow entering ground. (In SceneProperty we have `allowEnterGround: boolean`).
For me it makes more sense to keep that perspective here too and calculate whether can enter the ground or not. What do you think? In your code it matters more about keeping above the ground, but that is the default state so not sure...

src/components/molecules/Visualizer/Engine/Cesium/Box/Edge.tsx
keiya01: This indicates `cbp` is immutable. I want to create `CallbackProperty` only once. 
If we use `useMemo`, we need to set `deps array` to second args, and this will cause re-creating `CallbackProperty` and re-rendering. As a result, this will degrade performance.
But if you have any idea to improve this approach, I want to use it :eyes:

src/components/molecules/Visualizer/Plugin/context.tsx
keiya01: I fixed this review in [refactor: rename SelectedRearthEventType](https://github.com/reearth/reearth-web/pull/357/commits/7884c159efb0c928bce18cec9038d678589ec3dc)

src/components/molecules/Visualizer/Engine/Cesium/Box/utils.ts
keiya01: Sure. I moved this method to utils. [refactor: move translationWithClamping to Cesium/utils](https://github.com/reearth/reearth-web/pull/357/commits/166258bc1c306d90ef8a4998ed2f7131a85c1bfd)

src/components/molecules/Visualizer/Engine/Cesium/Box/utils.ts
keiya01: I renamed function name in [refactor: rename dotProductMousePosition](https://github.com/reearth/reearth-web/pull/357/commits/10dd8a5b3a56ee39e28596d2a1769d4e95bdea10).
Also about this function, it calculates mouse move amount on globe from mouse move amount from 2D screen.
I did a little refactor in [refactor: return moveAmount in dotMousePosition](https://github.com/reearth/reearth-web/pull/357/commits/5a6dbc2357aaef8f105ecec465501eb3b872947b).

src/components/molecules/Visualizer/Engine/Cesium/Box/utils.ts
keiya01: Exactly, I fixed this in [refactor: rename computeMoveAmount](https://github.com/reearth/reearth-web/pull/357/commits/f89ea5836794aa4d5a6a570da3ab4ae080462421)

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: I renamed this in [refactor: rename isProgressSamplingTerrainHeight](https://github.com/reearth/reearth-web/pull/357/commits/8b4ce98f4db663409909e18faa2ced0d7a418d56)

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: Thank you, this is my mistake. I fixed this in [fix: assigning inProgressSamplingTerrainHeight inside of IF statement](https://github.com/reearth/reearth-web/pull/357/commits/67e88b61fd94b8056fe6189788e62a4bf445571c).

src/util/value.ts
keiya01: I didn't know allowEnterGround property is exist. I fixed to use this property in [fix: use allowEnterGround](https://github.com/reearth/reearth-web/pull/357/commits/f70ee2e2ada51f9d1269bf37fbcc275671240a3a)

keiya01:
I did refactor to move hooks to hooks file in [refactor: separate component logic to hooks](https://github.com/reearth/reearth-web/pull/357/commits/5428128360303ac97f7f40248084630c09cfa99c)

keiya01:
@rot1024 
I fixed the box event to use layeredit event. [fix: use layeredit event to scale and rotate box](https://github.com/reearth/reearth-web/pull/357/commits/263622827313385c20752eb0b0fc2eb8686784cb)

src/components/molecules/Visualizer/Engine/Cesium/Box/Edge.tsx
KaWaite: Gotcha. Yeah the only other thing I can think of is just using useMemo with the `// eslint-disable-next-line react-hooks/exhaustive-deps` and then just have an empty deps array. But perhaps your approach is nicer since we don't have to disable anything.

But...do you actually not want `cbp` updated even if either `trs` or `edge` change?

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
KaWaite: Looks good! But you missed one point I think (or maybe it is on purpose??).
Where you use `updateTerrainHeight` is always inside of a `if(!allowEnterGround){}` statement. Then, inside of `updateTerrainHeight`, you check again if `!allowEnterGround`. One of the IF statements isn't necessary, right?

src/components/molecules/Visualizer/Engine/Cesium/Box/Edge.tsx
keiya01: Yes, I don't want to update `cbp` even if `trs` or `edge` is changed.
Because if `cbp` is updated, the entity component of recium is re-rendered. `trs` is changed too many times, so updating `cbp` occurs too many re-rendering in the entity component, then the screen is jaggy.

src/components/molecules/Visualizer/Engine/Cesium/Box/index.tsx
keiya01: Yeees!! I removed if statement in [refactor: remove unnecessary if statement](https://github.com/reearth/reearth-web/pull/357/commits/f69880028b5eeec6726f0ad51e4ee59a4e324769).

keiya01:
@KaWaite
I have done refactor to commonize some logic in [refactor: use sampleTerrainHeightFromCartesian](https://github.com/reearth/reearth-web/pull/357/commits/376a67868d7054a6a5e2ff97a8632c5e34a1199d)

KaWaite (approved):

rot1024 (approved):

PR #364, opened 2022-11-30
feat: extend plugin API supports communication

airslice:
close https://github.com/reearth/reearth/issues/376

# Overview

This PR enables plugin communication with other plugins.

## What I've done

Implement the API:

```ts
type PluginExtensionInstance = {
  id: string;
  pluginId: string;
  name: string; 
  extensionId: string;
  extensionType: "widget" | "block";
};

reearth.plugins.instances: PluginExtensionInstance[]
reearth.plugins.postMessage(id: string; message: any)
reearth.on("pluginmessage", (e: { data: any, sender: string }) => {})
```

## How I tested

Test with plugin:
[plugincomm.zip](https://github.com/reearth/reearth-web/files/10172887/plugincomm.zip)

Download & install & add multiple instances for test.
Also add block for test.

## Screenshot
localhost_3000_edit_01gjy7yqpqnjzcvj0wsk7h0f4w_preview(1440 desktop)


src/components/molecules/Visualizer/hooks.ts
```diff
@@ -257,6 +260,12 @@ export default ({
     alignSystem,
   });
 
+  const { pluginInstances } = usePluginInstances({
+    alignSystem,
+    floatingWidgets,
+    blocks: infobox?.blocks ?? selectedLayer?.infobox?.blocks,
```
KaWaite, line 266: Doesn't `infobox` get assigned from `selectedLayer?.infobox` in `useLayers`? If that is correct, we don't need the fallback because they are the same thing

src/components/molecules/Visualizer/usePluginInstances.ts
```diff
@@ -0,0 +1,118 @@
...
+  }, []);
+
+  const pluginInstances = useMemo(() => {
+    return {
+      meta: pluginInstancesMeta,
+      postMessage: pluginPostMessage,
+      addPluginMessageSender,
+      removePluginMessageSender,
+    };
+  }, [pluginInstancesMeta, pluginPostMessage, addPluginMessageSender, removePluginMessageSender]);
```
KaWaite, line 103: Why is this necessary instead of just returning them normally? I feel like this isn't necessary but guessing you did this for a specific reason.
But, all of the returned values are either useState or useCallback, which if they change should force a re-render.

src/components/molecules/Visualizer/hooks.ts
airslice: Yes you're right. Should not use `infobox?.blocks` here. Updated.

rot1024 (changes requested):

src/components/molecules/Visualizer/usePluginInstances.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const pluginMessageSenders = useRef<Map<string, (msg: any) => void>>(new Map());
+
+  const pluginInstances = useMemo(() => {
+    return {
+      meta: pluginInstancesMeta,
+      postMessage: (id: string, msg: any, sender: string) => {
+        const msgSender = pluginMessageSenders.current?.get(id);
+        if (!msgSender) return;
+        msgSender({ data: msg, sender });
```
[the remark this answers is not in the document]
airslice: Updated.

rot1024 (approved):

PR #369, opened 2022-12-05
feat: sandbox plugin iframe

airslice:
# Overview

For plugin iframes `about:srcdoc` meets the same-origin policy with parent, which will lead to unexpected access. I added `sandbox` property to iframe to make them now cross origin. At the same time the parent also can not get the iframe document.

## What I've done

- Add `sandbox` property to iframe. Currently allowed `allow-scripts` only.
- For `srcdoc`: use string html instead of empty string since we can't edit iframe document dynamically now.


src/components/atoms/Plugin/IFrame/hooks.ts
```diff
@@ -198,52 +135,36 @@ export default function useHook({
...
+    return `<script id="_reearth_resize">
+      let inited = false;
+      if ("ResizeObserver" in window) {         
+        new window.ResizeObserver(entries => {
+          const win = document.defaultView;
+          const html = document.body.parentElement;
+          const st = win.getComputedStyle(html, "");
+          horizontalMargin = parseInt(st.getPropertyValue("margin-left"), 10) + parseInt(st.getPropertyValue("margin-right"), 10);
+          verticalMargin = parseInt(st.getPropertyValue("margin-top"), 10) + parseInt(st.getPropertyValue("margin-bottom"), 10);
+          const width = html.offsetWidth + horizontalMargin;
```
airslice, line 148: Here i removed the conditions because the auto resize script will only inject once & should always get size from html.

src/components/atoms/Plugin/IFrame/index.tsx
```diff
@@ -63,11 +64,12 @@ const IFrame: React.ForwardRefRenderFunction<Ref, Props> = (
       frameBorder="no"
       scrolling={autoResize ? undefined : "no"}
       data-testid="iframe"
-      srcDoc=""
+      srcDoc={srcDoc}
       key={html}
       ref={composeRefs(iFrameRef, externalRef)}
       className={className}
       onLoad={onIFrameLoad}
+      sandbox="allow-scripts"
```
airslice, line 72: This may be too strict? I'm doing tests for exist plugins, maybe add some more if needed.

airslice:
This PR is out of date. There will be a new PR to implement this feature.

PR #375, opened 2022-12-14
fix: rendering bug in new layer system

keiya01:
# Overview

## What I've done

I fixed the bug that does not render the layer even when updating the property for layer.
This is because, the old system is updated every clock event is occurred but the new layer system disable this behavior for performance, so we need to call `scene.requestRender` manually.
Also I fixed resium in https://github.com/reearth/resium/pull/568.

~~TODO: I will update the resium when @rot1024 -san publish it.~~

## How I tested

1. Open storybook
2. Change the color of marker on storybook(That works)
3. Change the coordinates of marker on storybook(That didn't work but currently that should work)

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
         "react-spinners": "0.13.7",
         "react-use": "17.4.0",
         "remark-gfm": "3.0.1",
-        "resium": "1.15.1",
+        "resium": "^1.16.0",
         "tinycolor2": "1.4.2",
         "ts-easing": "0.2.0",
         "use-callback-ref": "1.3.0",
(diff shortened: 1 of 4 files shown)


rot1024 (approved):

package.json
```diff
@@ -152,7 +152,7 @@
         "react-spinners": "0.13.7",
         "react-use": "17.4.0",
         "remark-gfm": "3.0.1",
-        "resium": "1.15.1",
+        "resium": "^1.16.0",
```
KaWaite, line 155: Please make sure you don't add `^` to the version so all devs are using the exact version. If we don't, we sometimes get a single dev with an issue with their local environment that is hard to troubleshoot together, among other reasons!

KaWaite (approved):

package.json
keiya01: I fixed in https://github.com/reearth/reearth-web/pull/375/commits/457ee177070917b78cd8f1af5b427d6a4cd78b55.
Thank you.

PR #384, opened 2022-12-20
feat: reearth style language

pyshx:
# Overview

## What I've done
I've added support for style language to `reearth/core/mantle/evaluator`, we can support 
- Data Types in expressions
    - JSON → must support
        - boolean, number, string, null, array, object
- JSON Path support in variable statement to access nested attributes (properties)

## What I haven't done
I haven't added a support for vec1~4, Color, RegExp yet.

## How I tested
- Added unitTests using vitest.

- I have added `[key: string]: any; ` as one of the field for the features for testing purpose, If we know what it'll will look like in reality, I'd like add those fields and after that can also go ahead with adding the `define` support as well. https://github.com/reearth/reearth-web/blob/527fe328ea089876640f57b1319002848e7cf2d1/src/core/mantle/types/index.ts#L72


package.json
```diff
@@ -123,12 +123,15 @@
         "i18next-browser-languagedetector": "7.0.1",
         "jotai": "1.11.2",
         "js-file-download": "0.4.12",
+        "jsep": "^1.3.8",
+        "jsonpath-plus": "^7.2.0",
```
KaWaite, line 127: Please remove `^` from newly added deps. If we don't, local environments will be using different versions and it causes headaches trying to troubleshoot local issues.

src/core/mantle/evaluator/simple/expression.ts
```diff
@@ -0,0 +1,900 @@
+import { defined, Math as CesiumMath } from "cesium";
```
KaWaite, line 1: Is there any way you can split this file (expresssion.ts) in to multiple files? With over 700 lines I'm really hoping you can break it up in to smaller pieces somehow.

src/core/mantle/types/index.ts
```diff
@@ -70,12 +71,12 @@ export type DataRange = {
 export type DataType = "geojson" | "3dtiles";
 
 // Feature
-
 export type Feature = {
   id: string;
   geometry?: Geometry;
   properties?: any;
   range?: DataRange;
+  [key: string]: any;
```
rot1024, line 79:
```suggestion
```

Variable can refer to `id` and fields in `properties` of a Feature, not Feature itserlf.

src/core/mantle/evaluator/simple/expression.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      phoneNumbers: [
+        {
+          type: "iPhone",
+          number: "0123-4567-8888",
+        },
+        {
+          type: "home",
+          number: "0123-4567-8910",
+        },
+      ],
```
rot1024, line 81:
```suggestion
      properties: {
	    firstName: "John",
	    lastName: "doe",
	    age: 26,
	    address: {
	      streetAddress: "naist street",
	      city: "Nara",
	      postalCode: "630-0192",
	    },
	    phoneNumbers: [
	      {
	        type: "iPhone",
	        number: "0123-4567-8888",
	      },
	      {
	        type: "home",
	        number: "0123-4567-8910",
	      },
	    ],
      },
```

src/core/mantle/evaluator/simple/index.ts
```diff
@@ -32,17 +36,26 @@ export function evalLayerAppearances(
...
   );
 }
 
-function evalExpression(_e: Expression, _layer: LayerSimple, _feature?: Feature): unknown {
-  // TODO: eval
-  return undefined;
+function evalExpression(
+  _styleExpression: StyleExpression,
+  _layer: LayerSimple,
+  _feature?: Feature,
```
rot1024, line 49:
```suggestion
  styleExpression: StyleExpression,
  layer: LayerSimple,
  feature?: Feature,
```

I use `_` to supress "unused arguments" warning from TypeScript, but they are used.

rot1024 (changes requested):

src/core/mantle/types/expression.ts
```diff
@@ -0,0 +1,13 @@
...
+  conditions: [string, string][];
+};
+
+export type UnaryFunction = (call: string, left: number) => number;
+export type BinaryFunction = (call: string, left: number, right: number) => number;
+
+export type JPLiteral = {
+  literalName: string;
+  literalValue: any;
+};
```
rot1024, line 13:
```suggestion
```

They are only used internally, so move them to `expression.ts`

rot1024 (changes requested):

src/core/mantle/evaluator/simple/expression.ts
pyshx: I can, but I was thinking its better to have everything associated to the evaluation part together as the components inside this file aren't gonna be used anywhere else, what do you think?

src/core/mantle/types/expression.ts
```diff
@@ -1,3 +1,7 @@
+export type ExpressionContainer = {
+  expression: StyleExpression;
+};
+
 export type StyleExpression = ConditionsExpression | string | boolean | number;
```
rot1024, line 5:
```suggestion
export type StyleExpression = ConditionsExpression | string;
```

Always an expression should be described as a string?

rot1024 (approved):

src/core/mantle/types/expression.ts
pyshx: Yeah, To enforce the expression behaviour.
Though I went with 
```
{
  marker: {
    pointSize: 23,
  }
}
```

```
{
 marker: {
   pointSize: {
     expression: 23,
     }
   }
}
```
Both valid statements.

rot1024: Oh ok, but any type is settable as a value (`T[K] | StyleExpression`). so to disambiguate, it would be better to allow only expressions to be written inside EXPRESSION. If you want to embed a value directly, write it without expression.

pyshx: makes sense, done.

KaWaite (approved):

PR #386, opened 2022-12-21
feat: support glTF on reearth/core

keiya01:
# Overview

## What I've done

I've supported glTF file on the reearth/core.

## Screenshot

Screen Shot 2022-12-21 at 17 28 17

### DIFF
--- a/src/core/engines/Cesium/Feature/Model/index.stories.tsx
+++ b/src/core/engines/Cesium/Feature/Model/index.stories.tsx
@@ -33,7 +33,7 @@ Default.args = {
         },
       },
       model: {
-        model: "/BoxAnimated.glb",
+        url: "/BoxAnimated.glb",
         scale: 1000000,
       },
     },
--- a/src/core/engines/Cesium/Feature/Model/index.stories.tsx
+++ b/src/core/engines/Cesium/Feature/Model/index.stories.tsx
@@ -1,0 +1,49 @@
+import { Meta, Story } from "@storybook/react";
+
+import { engine } from "../../";
+import Component, { Props } from "../../../../Map";
+
+export default {
+  title: "core/engines/Cesium/Feature/Model",
+  component: Component,
+  parameters: { actions: { argTypesRegex: "^on.*" } },
+} as Meta;
+
+const Template: Story<Props> = args => <Component {...args} />;
+
+export const Default = Template.bind([]);
+Default.args = {
+  engine: "cesium",
+  engines: {
+    cesium: engine,
+  },
+  ready: true,
+  layers: [
+    {
+      id: "l",
+      type: "simple",
+      data: {
+        type: "geojson",
+        value: {
+          type: "Feature",
+          geometry: {
+            type: "Point",
+            coordinates: [0, 0, 1000],
+          },
+        },
+      },
+      model: {
+        model: "/BoxAnimated.glb",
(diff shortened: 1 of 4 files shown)


src/core/engines/Cesium/Feature/Model/index.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+        value: {
+          type: "Feature",
+          geometry: {
+            type: "Point",
+            coordinates: [0, 0, 1000],
+          },
+        },
+      },
+      model: {
+        model: "/BoxAnimated.glb",
```
rot1024, line 36:
```suggestion
        url: "/BoxAnimated.glb",
```

Can you support `url` property also?  Set `model={url || model}` in `Feature/Model/index.tsx`

rot1024 (approved):

src/core/mantle/types/appearance.ts
```diff
@@ -84,7 +84,9 @@ export type EllipsoidAppearance = {
 };
 
 export type ModelAppearance = {
+  // For compat
```
KaWaite, line 87: Is the whole ModelAppearance type "for compatiblility"? Or just the model prop? If just the model prop, move this to that line (like the `// default: true` below on line 97) so that it is clearer.

src/core/engines/Cesium/Feature/Model/index.tsx
```diff
@@ -78,7 +79,7 @@ export default function Model({ id, isVisible, property, geometry, layer, featur
       featureId={feature?.id}
       draggable>
       <ModelGraphics
-        uri={model}
+        uri={model || url}
```
KaWaite, line 82: The below is just something to keep in mind as you develop. I am assuming the user will be able to choose a TYPE (model or URL) from the right panel. If that isn't the case, ignore the rest of this message.

I've seen issues before while using Re:Earth that are caused by code like this. When a user can choose a type (ie. marker can be icon or point, this Model can be model or url, etc), if the user starts with model, and then switches to URL, or vice versa, both fields will have a value in the DB. Which means, even if the user selects a url type, if there is a model it'll show that.

Because of this, keep in mind that when we query for the model, we will need to either:

- Assign null/undefined after querying based on selected type(aka. If the type is url, make model null/undefined, if the type is model, make the url null/undefined.
OR
- Make sure the query conditionally fetches the values (see [directives](https://www.apollographql.com/docs/apollo-server/schema/directives/))

Maybe you already know of this, but if not..!

KaWaite (approved):

src/core/engines/Cesium/Feature/Model/index.tsx
keiya01: Thank you.
I think `model` props will be only URL actually. The props was used for specifying URL but this name makes confusing, so I want to integrate the name from `model` to `url`. But current users are using `model` props so we need to support `model` too in same way with `url`:pray:

src/core/mantle/types/appearance.ts
keiya01: https://github.com/reearth/reearth-web/pull/387/commits/9cfe5d7c35485681e0a37a690859e049bf3425ec
I'm sorry I ended up merging this PR unintentionally, so I fixed this comment in another PR...

PR #396, opened 2023-01-13
fix: cache ComputedFeature on reearth/core

keiya01:
# Overview

I fixed to cache computed feature on reearth/core.

## What I've done

- added `writeComputedFeature` to cache computed feature directly.
- fixed some logic in Tileset feature component.
- fixed some logic in Raster feature component for MVT.

## What I haven't done

- I don't add feature to select feature.


src/core/mantle/atoms/compute.ts
```diff
@@ -157,6 +142,45 @@ export function computeAtom(cache?: typeof globalDataFeaturesCache) {
     await set(compute, undefined);
   });
 
+  const writeComputedFeatures = atom(null, async (get, set, value: ComputedFeature[]) => {
```
keiya01, line 145: We can use `writeFeatures` atom, but delegated data is evaluated before invoking `writeFeatures`.
So I think we can skip the process, and we should use `writeComputedFeatures` to cache ComputedFeature directly.
Do you think about this? @rot1024

src/core/mantle/atoms/compute.ts
```diff
@@ -157,6 +142,48 @@ export function computeAtom(cache?: typeof globalDataFeaturesCache) {
     await set(compute, undefined);
   });
 
+  const writeComputedFeatures = atom(
+    null,
+    async (get, set, value: { feature: Feature[]; computed: ComputedFeature[] }) => {
+      const currentLayer = get(layer);
+      if (currentLayer?.type !== "simple" || !currentLayer.data) return;
+
+      if (process.env.NODE_ENV === "development") {
```
rot1024, line 151:
```suggestion
      if (import.meta.env.DEV) {
```

https://ja.vitejs.dev/guide/env-and-mode.html

rot1024 (approved):

PR #408, opened 2023-01-26
feat: support general transit feed

pyshx:
# Overview
Add support for transit feed


src/core/mantle/data/gtfs.ts
```diff
@@ -0,0 +1,126 @@
...
+      id: f.id,
+      geometry: {
+        type: "Point",
+        coordinates: [f.properties.position?.longitude || 0, f.properties.position?.latitude || 0],
+      },
+      properties: {},
+    });
+  });
+
+  return result;
```
rot1024, line 59:
```suggestion
  return tripsData?.trips?.filter(f => !!f.properties.position).map(f => {
    return {
      type: "feature",
      id: f.id,
      geometry: {
        type: "Point",
        coordinates: [f.properties.position?.longitude || 0, f.properties.position?.latitude || 0],
      },
      properties: {},
    };
  });
```

src/core/mantle/data/gtfs.ts
```diff
@@ -0,0 +1,126 @@
...
+//     case OccupancyStatus.FULL:
+//       return "#cf222e";
+//     case OccupancyStatus.STANDING_ROOM_ONLY:
+//       return "#dc6d1a";
+//     case OccupancyStatus.FEW_SEATS_AVAILABLE:
+//       return "#f6e05e";
+//     default:
+//       return "#0969da";
+//   }
+// };
```
rot1024, line 126:
```suggestion
```

Style is completely left to the users.

src/core/mantle/data/gtfs.ts
```diff
@@ -0,0 +1,126 @@
+import Pbf from "pbf";
+
+import type { Data, DataRange, Feature } from "../types";
+
+import { Trips, Trip, GTFS, GTFSReader } from "./gtfsReader";
+// import { OccupancyStatus } from "./gtfsReader";
```
rot1024, line 6:
```suggestion
```

src/core/mantle/types/appearance.ts
```diff
@@ -101,6 +101,7 @@ export type ModelAppearance = {
   lightColor?: string;
   silhouette?: boolean;
   silhouetteColor?: string;
+  bearingProperty?: number;
```
rot1024, line 104:
```suggestion
  bearing?: number;
```

src/core/mantle/types/index.ts
```diff
@@ -60,6 +60,7 @@ export type NaiveBlock<P = any> = Omit<Block<P>, "id">;
 export type Data = {
   type: DataType;
   url?: string;
+  refreshInterval?: number;
```
rot1024, line 63:
```suggestion
```

rot1024 (changes requested):

src/core/mantle/data/gtfs.ts
pyshx: OK.

src/core/mantle/data/gtfs.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  tripsData?.trips?.flatMap(f => {
+    result.push({
+      type: "feature",
+      id: f.id,
+      geometry: {
+        type: "Point",
+        coordinates: [f.properties.position?.longitude || 0, f.properties.position?.latitude || 0],
+      },
+      properties: {},
```
rot1024, line 55: bearing seems not injected

rot1024 (changes requested):

rot1024 (approved):

src/core/mantle/data/gtfs.ts
```diff
@@ -0,0 +1,112 @@
+import Pbf from "pbf";
+
+import type { Data, DataRange, Feature } from "../types";
+
+import { Trips, Trip, GTFS, GTFSReader } from "./gtfsReader";
+import { f } from "./utils";
+
+let gtfs: GTFS = {};
```
keiya01, line 8: Why is this value global? I think it's better to close within local scope.

src/core/mantle/data/gtfsReader.ts
```diff
@@ -0,0 +1,454 @@
...
+}
+
+export class TimeRangeReader extends ProtobufMessageReader<TimeRange> {
+  readonly defaultMessage: TimeRange = {
+    start: 0,
+    end: 0,
+  };
+
+  protected readField(tag: number, obj = {} as TimeRange, pbf: Pbf): void {
+    if (tag === 1) obj.start = pbf.readVarint();
```
keiya01, line 356: IMO: Maybe it's better to attach specific name for these number.

src/core/mantle/data/gtfs.ts
pyshx: It is but I left it like that for interval fetching where the values will be refreshed in every cycle.

Take a look at this
https://github.com/reearth/reearth-web/pull/428/commits

keiya01: I see. It looks good.

src/core/mantle/data/gtfsReader.ts
pyshx: It could've been like that if all of them were part of common context(say 1 mean int, 2 mean float) but here each `readField` call has used them for different members of the corresponding class.

Since those const wouldn't have been used by other files, I decided to make it smaller.

keiya01: It makes sense.

keiya01 (approved):

PR #418, opened 2023-02-02
fix: layer appearances are not evaluated as expected

pyshx:
# Overview
This PR fixes a BUG observed and a small refactor.

## What I've done
I have passed one of the features into evalLayerAppearances function for providing a feature reference to layerAppearance evaluation.

## How I tested
I've test this fix on my local instance against 
```
reearth.layers.add({
  // always same
  type: "simple",

  // 1. how to load data
  data: { 
    type: "geojson",
    value: {
		  "type": "FeatureCollection",
		  "features": [
		    {
		      "type": "Feature",
		      "properties": {
            "firstName": "John",
            "lastName": "doe",
            "age": 26,
            "address": {
                "streetAddress": "naist street",
                "city": "Nara",
                "postalCode": "630-0192"
            },
            "phoneNumbers": [
                {
                    "type": "iPhone",
                    "number": "0123-4567-8888"
                },
                {
                    "type": "home",
                    "number": "0123-4567-8910"
                }
            ]
          },
		      "geometry": {
		        "coordinates": [
		          -15.209829106984472,
		          20.323569554406248
		        ],
		        "type": "Point"
		      }
		    },
		    {
		      "type": "Feature",
		      "properties": {
            "firstName": "John",
            "lastName": "doe",
            "age": 26,
            "address": {
                "streetAddress": "naist street",
                "city": "Nara",
                "postalCode": "630-0192"
            },
            "phoneNumbers": [
                {
                    "type": "iPhone!",
                    "number": "0123-4567-8888"
                },
                {
                    "type": "home",
                    "number": "0123-4567-8910"
                }
            ]
          },
		      "geometry": {
		        "coordinates": [
		          -8.061509150472233,
		          21.512505329690285
		        ],
		        "type": "Point"
		      }
		    }
		  ]
		}
  },

  // 2. how to style
  marker: {
    imageColor: {
      expression: '${$.phoneNumbers[:1].type} === "iPhone"?color("blue"):color("red")'
    }
  }
});
```

Screenshot 2023-02-02 at 18 48 12

## Screenshot of the bug

Screenshot 2023-02-02 at 18 45 04


src/core/mantle/evaluator/simple/expression/utils.ts
```diff
@@ -0,0 +1 @@
+export { default as generateRandomString } from "@reearth/util/generate-random-string";
```
rot1024, line 1: Can you copy implmentation here?

src/core/mantle/evaluator/simple/index.ts
```diff
@@ -20,9 +20,10 @@ export async function evalSimpleLayer(
   ctx: EvalContext,
 ): Promise<EvalResult | undefined> {
   const features = layer.data ? await ctx.getAllFeatures(layer.data) : undefined;
+  const firstFeature = [...(features || [])].shift();
   const appearances: Partial<LayerAppearanceTypes> = pick(layer, appearanceKeys);
   return {
-    layer: evalLayerAppearances(appearances, layer),
+    layer: evalLayerAppearances(appearances, layer, firstFeature),
```
rot1024, line 26:
```suggestion
    layer: evalLayerAppearances(appearances, laye),
```

This is probably inappropriate. I think the problem lies at this line:

```ts
features?.map(f => evalSimpleLayerFeature(layer, f)),
```

src/core/mantle/evaluator/simple/index.ts
```diff
@@ -20,9 +20,10 @@ export async function evalSimpleLayer(
   ctx: EvalContext,
 ): Promise<EvalResult | undefined> {
   const features = layer.data ? await ctx.getAllFeatures(layer.data) : undefined;
+  const firstFeature = [...(features || [])].shift();
```
rot1024, line 23:
```suggestion
```

rot1024 (changes requested):

src/core/mantle/evaluator/simple/expression/utils.ts
pyshx: Done.

src/core/mantle/evaluator/simple/index.ts
pyshx: Not really, the problem is that we are calling `evalLayerApperances` without any features in it and VariableReplacer doesn't like it because if the used expression is JSONPATH variableReplacer demands features(properties) which wouldn't be provided in the above example. That's the cause of this BUG.

pyshx: But I agree with you on the inappropriate part.

src/core/mantle/evaluator/simple/index.ts
```diff
@@ -42,6 +42,17 @@ export function evalLayerAppearances(
...
+    if (layer.properties && layer.id) {
+      feature = {
+        type: "feature",
+        id: layer.id,
+        properties: layer.properties,
+      };
+    } else {
+      throw new Error(`layer.properties and layer.id is required`);
+    }
+  }
```
rot1024, line 55:
```suggestion
  if (!feature) {
    if (!layer.id) {
      throw new Error("layer id is required");
    }
    feature = {
      type: "feature",
      id: layer.id,
      properties: layer.properties || {},
    };
  }
```

rot1024 (changes requested):

rot1024 (approved):

PR #432, opened 2023-02-07
fix: GeoJSON with resource appearance on reearth/core

keiya01:
# Overview

Resource feature component load geojson even if geojson is already rendered by other appearance like marker.
So I fixed this bugs.


keiya01:
These should work.

```
reearth.layers.add({
  type: "simple",
  data: {
    type: "geojson",
    url: "FUGA"
  },
  marker: {},
});
```

```
reearth.layers.add({
  type: "simple",
  data: {
    type: "geojson",
    value: {},
  },
  resource: {
    url: "HOGE"
  },
});

```

rot1024 (approved):

PR #435, opened 2023-02-07
fix: support resource auto on reearth/core

keiya01:
# Overview

## Screenshot
Screen Shot 2023-02-07 at 19 37 20

### DIFF
--- a/src/util/path.ts
+++ b/src/util/path.ts
@@ -11,5 +11,10 @@ export const parseHost = (source: DataSource): string | undefined => {
   return parsed[2];
 };
 
-export const getExtname = (url: string | undefined): string | undefined =>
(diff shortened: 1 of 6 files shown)


src/util/path.ts
```diff
@@ -10,3 +10,6 @@ export const parseHost = (source: DataSource): string | undefined => {
 
   return parsed[2];
 };
+
+export const getExtname = (url: string | undefined): string | undefined =>
+  url?.match(/\.([a-z]+?)(?:\?.*?)?$/)?.[1];
```
rot1024, line 15:
```suggestion
export const getExtname = (url: string | undefined): string | undefined => {
  try {
    return new URL(url).pathname.split(".")[1];
  } catch {
    return undefined;
  }
};
```

rot1024 (approved):

PR #470, opened 2023-02-17
feat: layer select event in reearth/core

rot1024:
```js
reearth.layers.add({
  type: "simple",
  data: {
    type: "geojson",
    value: {
      type: "Feature",
      geometry: {
        type: "Point",
        coordinates: [0, 0]
      },
      properties: {
        url: "https://example.com"
      }
    }
  },
  marker: {
    imageSize: 2
  },
  events: {
    select: {
      openUrl: {
        urlKey: "url"
      }
    }
  }
})
```


src/core/Visualizer/hooks.ts
```diff
@@ -112,6 +114,9 @@ export default function useHooks({
     },
     [],
   );
+  useEffect(() => {
+    handleLayerSelect(selectedLayerId?.layerId, selectedLayerId?.featureId, undefined, undefined);
```
rot1024, line 118: It's ok? (third arg is undefined)

keiya01 (approved):

keiya01 (approved):

PR #474, opened 2023-02-20
fix: color function on reearth/core

keiya01:
# Overview


pyshx (approved):
Thanks for the fixing my blunder.

rot1024 (approved):

PR #518, opened 2023-03-07
feat: find feature API on reearth/core

keiya01:
# Overview

Using `reearth.layers.layers[number].computed` makes application be heavy, so I added `findFeature` APIs.


src/core/Map/Layers/utils.ts
```diff
@@ -50,12 +50,13 @@ export const copyLazyLayer = (l: LazyLayer | undefined) => {
...
-            return computedRes;
-          }, {} as Record<string, any>);
-        },
+        // get [key]() {
+        //   return computedLayerKeys.reduce((computedRes, computedKey) => {
+        //     computedRes[computedKey] = l?.[key as keyof LazyLayer]?.[computedKey];
+        //     return computedRes;
+        //   }, {} as Record<string, any>);
+        // },
+        [key]: l?.[key], // For debug. We can not access to `computed` on plugin.
```
keiya01, line 59: `computed` is cloned(or copy) deeply on quickjs runtime, so it makes application be heavy.

rot1024 (approved):

PR #544, opened 2023-03-14
fix: draw polylines on polygon on reearth/core

rot1024:


src/core/engines/Cesium/Feature/Polygon/index.tsx
```diff
@@ -17,13 +17,21 @@ import {
   type FeatureProps,
 } from "../utils";
 
-export type Props = FeatureProps<Property>;
+export type Props = FeatureProps<Property> & { disableWorkaround?: boolean };
```
keiya01, line 20: When do we use this option?

keiya01 (approved):
LGTM!

src/core/engines/Cesium/Feature/Polygon/index.tsx
rot1024: It is mostly for debugging purposes as there is no opportunity to use it. When deleting a workaround someday, the implementation will make it easier to see where the workaround should be deleted.

PR #562, opened 2023-03-20
fix: infobox html block styling

issmail-basel:
# Overview

# To be styled
- Font size
- Font family
- Color (already styled, maybe)
- Link Color (it should be an accent color of the theme)?

## Screenshot
Screenshot 2023-03-20 at 8 41 38 PM
Screenshot 2023-03-20 at 8 41 27 PM

### DIFF
--- a/src/components/molecules/Visualizer/Block/HTML/index.tsx
+++ b/src/components/molecules/Visualizer/Block/HTML/index.tsx
@@ -105,8 +105,22 @@ const HTMLBlock: React.FC<Props> = ({
 
     // Initialize styles
     frameWindow.document.documentElement.style.margin = "0";
-    frameWindow.document.body.style.color = themeColor ?? getComputedStyle(frameRef).color;
-    frameWindow.document.body.style.margin = "0";
+
+    // Check if a style element has already been appended to the head
+    let style: HTMLElement | null = frameWindow.document.querySelector(
+      'style[data-id="iframe-style"]',
+    );
+    if (!style) {
+      // Create a new style element if it doesn't exist
+      style = frameWindow.document.createElement("style");
+      style.dataset.id = "iframe-style";
+      frameWindow.document.head.append(style);
+    }
+    // Update the content of the existing or new style element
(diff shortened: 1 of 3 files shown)


yarn.lock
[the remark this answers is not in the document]
keiya01: I think it's no problem. I forgot removing `^` from yarn.lock:pray:

src/components/molecules/Visualizer/Block/HTML/index.tsx
```diff
@@ -105,8 +105,22 @@ const HTMLBlock: React.FC<Props> = ({
...
-    frameWindow.document.body.style.margin = "0";
+
+    // Check if a style element has already been appended to the head
+    let style: HTMLElement | null = frameWindow.document.querySelector(
+      'style[data-id="iframe-style"]',
+    );
+    if (!style) {
+      // Create a new style element if it doesn't exist
+      style = frameWindow.document.createElement("style");
+      style.dataset.id = "iframe-style";
```
keiya01, line 116:
```suggestion
      style.dataset.id = "reearth-iframe-style";
```
I prefer using `reearth` prefix because html may have same name.
And it's useful when we debug it because we can find this attributes is added by our code.

src/components/molecules/Visualizer/Block/HTML/index.tsx
```diff
@@ -105,8 +105,22 @@ const HTMLBlock: React.FC<Props> = ({
 
     // Initialize styles
     frameWindow.document.documentElement.style.margin = "0";
-    frameWindow.document.body.style.color = themeColor ?? getComputedStyle(frameRef).color;
-    frameWindow.document.body.style.margin = "0";
+
+    // Check if a style element has already been appended to the head
+    let style: HTMLElement | null = frameWindow.document.querySelector(
+      'style[data-id="iframe-style"]',
```
keiya01, line 111:
```suggestion
      'style[data-id="reearth-iframe-style"]',
```

keiya01 (approved):

PR #571, opened 2023-03-24
fix: timeline scroll should be fixed in initial render

keiya01:
# Overview

### DIFF
--- a/src/components/atoms/Timeline/hooks.ts
+++ b/src/components/atoms/Timeline/hooks.ts
@@ -123,7 +123,6 @@ const useTimelineInteraction = ({
 
     const obs = new ResizeObserver(m => {
       const target = m[0].target;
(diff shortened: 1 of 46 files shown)


src/components/atoms/Timeline/hooks.ts
```diff
@@ -110,19 +110,31 @@ const useTimelineInteraction = ({
...
+    [zoom, setZoom],
   );
 
+  useEffect(() => {
+    const elm = scaleElement.current;
+    if (!elm) return;
+
+    const obs = new ResizeObserver(m => {
+      const target = m[0].target;
+      console.log("First call???", m[0], target.clientWidth);
```
rot1024, line 128:
```suggestion
```

rot1024 (approved):

PR #572, opened 2023-03-24
fix: JSON Path condition on reearth/core

keiya01:
# Overview

### DIFF
--- a/src/core/mantle/evaluator/simple/expression/variableReplacer.ts
+++ b/src/core/mantle/evaluator/simple/expression/variableReplacer.ts
@@ -34,7 +34,7 @@ export function replaceVariables(expression: string, feature?: any): [string, JP
             return [result, []];
           }
         }
-        if (res.length == 1) {
+        if (res.length !== 0) {
           const placeholderLiteral = generateRandomString(10);
           literalJP.push({
             literalName: placeholderLiteral,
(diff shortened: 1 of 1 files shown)


src/core/mantle/evaluator/simple/expression/variableReplacer.ts
```diff
@@ -34,15 +34,15 @@ export function replaceVariables(expression: string, feature?: any): [string, JP
             return [result, []];
           }
         }
-        if (res.length == 1) {
+        if (res.length !== 0) {
```
keiya01, line 37: `res` length is not alway `1`, so I fixed condition.

src/core/mantle/evaluator/simple/expression/variableReplacer.ts
```diff
@@ -34,15 +34,15 @@ export function replaceVariables(expression: string, feature?: any): [string, JP
...
           const placeholderLiteral = generateRandomString(10);
           literalJP.push({
             literalName: placeholderLiteral,
             literalValue: res[0],
           });
           result += placeholderLiteral;
         } else {
-          return [result, []];
+          return ["false", []];
```
keiya01, line 45: Returning result stops evaluating other condition, so I fixed to return false.

src/core/mantle/evaluator/simple/expression/variableReplacer.ts
pyshx: That's intended, Because As of now we cannot handle multiple output of a JSONPath evaluation.
And its suggested to only use JSONPath which gives single output for now.

pyshx (approved):

src/core/mantle/evaluator/simple/expression/variableReplacer.ts
pyshx: Agreed.

PR #573, opened 2023-03-24
perf: add concurrency to evaluator and cache support

pyshx:
# Overview
This is WIP PR for improving styling performance of reearth/core.

### DIFF
--- a/src/core/mantle/evaluator/simple/index.ts
+++ b/src/core/mantle/evaluator/simple/index.ts
@@ -10,11 +10,14 @@ import {
   LayerSimple,
   ExpressionContainer,
   TimeInterval,
+  StyleExpression,
 } from "../../types";
 
 import { ConditionalExpression } from "./conditionalExpression";
 import { clearExpressionCaches, Expression } from "./expression";
 import { evalTimeInterval } from "./interval";
+
+// const EVAL_EXPRESSION_CACHES: Map<string, any> = new Map();
 
 export async function evalSimpleLayer(
   layer: LayerSimple,
@@ -36,11 +39,14 @@ export async function evalSimpleLayer(
   };
 }
 
-function evalFeaturesInWorker(
+async function evalFeaturesInWorker(
   features: Feature[],
   layer: LayerSimple,
   timeIntervals: TimeInterval[] | void,
 ): Promise<ComputedFeature[]> {
+  const numWorkers = navigator.hardwareConcurrency || 1;
+  const chunkSize = Math.ceil(features.length / numWorkers);
+
   const workerScript = `
     self.addEventListener('message', event => {
       const { features, layer, timeIntervals } = event.data;
@@ -55,17 +61,28 @@ function evalFeaturesInWorker(
   const workerBlob = new Blob([workerScript], { type: "text/javascript" });
   const workerUrl = URL.createObjectURL(workerBlob);
 
-  return new Promise(resolve => {
-    const worker = new Worker(workerUrl);
+  const workers = new Array(numWorkers);
+  const promises = new Array(numWorkers);
 
-    worker.addEventListener("message", event => {
-      const featureResults = event.data;
-      worker.terminate();
-      resolve(featureResults);
+  for (let i = 0; i < numWorkers; i++) {
+    const start = i * chunkSize;
+    const end = Math.min((i + 1) * chunkSize, features.length);
+    const chunk = features.slice(start, end);
+
+    workers[i] = new Worker(workerUrl);
+    promises[i] = new Promise(resolve => {
+      workers[i].addEventListener("message", (event: { data: any }) => {
+        const featureResults = event.data;
+        workers[i].terminate();
+        resolve(featureResults);
+      });
     });
 
-    worker.postMessage({ features, layer, timeIntervals });
-  });
+    workers[i].postMessage({ features: chunk, layer, timeIntervals });
+  }
+
+  const results = await Promise.all(promises);
+  return Array.prototype.concat(...results);
 }
 
 export const evalSimpleLayerFeature = (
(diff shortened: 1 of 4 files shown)


src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    const references: string[] = [];
+    for (const [condition, value] of expression.conditions) {
+      references.push(...getReferences(condition), ...getReferences(value));
+    }
+    return references;
+  }
+}
+
+export function getReferences(expression: string): string[] {
+  if (hasJsonPath(expression)) return ["REEARTH_JSONPATH"];
```
keiya01, line 21: IMO
It's better to use constant variable to avoid a mistake like typo.

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,67 @@
+import { pick } from "lodash-es";
+
+import { Feature, StyleExpression } from "../../types";
+
+const JSONPATH_REGEX =
+  /\$[a-zA-Z_][a-zA-Z0-9_]*(\[[^\]]+\])?(\.[a-zA-Z_][a-zA-Z0-9_]*(\[[^\]]+\])?)*$/g;
```
keiya01, line 6: I think it's better to write a test about regex for JSONPath.
Because I tried to test this regex with `$.phoneNumbers[:1].type` in https://regex101.com/, but it doesn't work.

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  } else {
+    const references: string[] = [];
+    for (const [condition, value] of expression.conditions) {
+      references.push(...getReferences(condition), ...getReferences(value));
+    }
+    return references;
+  }
+}
+
+export function getReferences(expression: string): string[] {
```
keiya01, line 20: I think this function is very heavy because this does same thing with replaceVariable.
So I think it's better to cache the result in here too like [I did](https://github.com/reearth/reearth-web/pull/574/files#diff-b985f02263b6110afa5cbeb5d637e627de98028bbcc80ed50c30ccbb2795980c).

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+    } else if (openDoubleQuote >= 0 && openDoubleQuote < i) {
+      const closeQuote = exp.indexOf('"', openDoubleQuote + 1);
+      result.push(exp.substring(0, closeQuote + 1));
+      exp = exp.substring(closeQuote + 1);
+    } else {
+      const j = exp.indexOf("}", i);
+
+      if (j < 0) {
+        console.log("Unmatched {.");
+      }
```
keiya01, line 47: Is this line necessary?

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+      const varExp = exp.slice(i + 2, j);
+      if (varExpRegex.test(varExp)) {
+        let res = jsonPathCache[varExp];
+        if (!res) {
+          try {
+            res = JSONPath({ json: feature, path: varExp });
+            jsonPathCache[varExp] = res;
+          } catch (error) {
+            console.log("Invalid JSONPath");
```
keiya01, line 56:
```suggestion
            console.error("Invalid JSONPath");
```

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+          }
+        }
+        if (res.length !== 0) {
+          console.log("JSONPathEval: ", res[0]);
```
keiya01, line 60:
```suggestion
```

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    i = exp.indexOf("${");
+  }
+
+  return result;
+}
+
+function getObjectPathByValue(obj: any, value: any): string | undefined {
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
```
keiya01, line 79:
```suggestion
    if (Object.hasOwn(obj, key)) {
```
I think hasOwn is better than it.

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+  } else {
+    const references: string[] = [];
+    for (const [condition, value] of expression.conditions) {
+      references.push(...getReferences(condition, feature), ...getReferences(value, feature));
+    }
+    return references;
+  }
+}
+
+export function getReferences(expression: string, feature?: any): string[] {
```
keiya01, line 23: We have two function to parse expression. First is this function. Second is `replaceVariables`.
But these are very heavy task, so it will make app slower...
I think it's better to share the result of parsing with Expression class, and if we remove replaceVariables then performance will be more improved.

src/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+          try {
+            res = JSONPath({ json: feature, path: varExp });
+            jsonPathCache[varExp] = res;
+          } catch (error) {
+            console.log("Invalid JSONPath");
+          }
+        }
+        if (res.length !== 0) {
+          console.log("JSONPathEval: ", res[0]);
+          const keyPath = getObjectPathByValue(feature, res[0]);
```
keiya01, line 61: I think JSONPath returns a only evaluated primitive value or object.
So if the result value is same with other field value, does it work?

For example, I assume that feature has the following property.
```
{
  count: 1,
  number: 1,
}
```

And if we write a condition like `["${$.number} === 1", "color('red')"]` then it returns `1`, but keyPath may be `[count]`, right?
And if we change the condition to `["${$.count} === 1", "color('blue')"]` and feature is same, the result will be cached value. Because JSONPath returns same `1` and `keyPath` will be `[count]`.

src/core/mantle/evaluator/simple/index.ts
```diff
@@ -111,11 +120,41 @@ function evalExpression(
       if (typeof styleExpression === "undefined") {
         return undefined;
       } else if (typeof styleExpression === "object" && styleExpression.conditions) {
-        return new ConditionalExpression(styleExpression, feature, layer.defines).evaluate();
+        const cacheKey = JSON.stringify([
+          styleExpression,
+          getCacheableProperties(styleExpression, feature?.properties),
```
keiya01, line 125: We don't have to instantiate Expression no more, but we became to need to parse expression every time...

src/core/mantle/evaluator/simple/utils.ts
pyshx: I looks it'll be hard to separate out alot the repetitive logic and avoid iterating over the string again in `replaceVariables` now that i've removed JP related evaluations.

src/core/mantle/evaluator/simple/index.ts
pyshx: Hopefully this will be less heavy with the cache for getReferences in place.

src/core/mantle/evaluator/simple/utils.ts
pyshx: Done.

keiya01 (approved):

keiya01 (approved):
Looks great! Thank you.

PR #574, opened 2023-03-25
perf: improve styling in MVT on reearth/core

keiya01:
# Overview

I did three things to improve performance.

1. Cache MVT tile in MVT lib.
2. Cache evaluated style in Raster component
3. Cache replacedVariable without JSONPath.

### DIFF
--- a/src/core/engines/Cesium/Feature/Raster/hooks.ts
+++ b/src/core/engines/Cesium/Feature/Raster/hooks.ts
@@ -137,6 +137,7 @@ export const useMVT = ({
   const imageryProvider = useMemo(() => {
     if (!isVisible || !show || !url || !layers || type !== "mvt") return;
     let currentTime: number | undefined;
+    const cachedStyleMap = new Map();
     return new MVTImageryProvider({
       minimumLevel,
       maximumLevel,
@@ -150,6 +151,11 @@ export const useMVT = ({
         return currentTime === updatedAt.current;
       },
       style: (mvtFeature, tile) => {
+        const styleCacheKey = JSON.stringify(mvtFeature.properties);
+        const cachedStyle = cachedStyleMap.get(styleCacheKey);
+        if (cachedStyle) {
+          return cachedStyle;
+        }
         const computedFeature = ((): ComputedFeature | void => {
           const layer = cachedCalculatedLayerRef.current?.layer;
           if (layer?.type === "simple" && VectorTileFeature.types[mvtFeature.type] === "Polygon") {
@@ -167,7 +173,7 @@ export const useMVT = ({
         })();
 
         const polygon = computedFeature?.polygon;
-        return {
+        const style = {
           fillStyle:
             (polygon?.fill ?? true) && (polygon?.show ?? true)
               ? polygon?.fillColor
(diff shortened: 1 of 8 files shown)


src/core/engines/Cesium/Feature/Raster/hooks.ts
```diff
@@ -150,6 +151,11 @@ export const useMVT = ({
         return currentTime === updatedAt.current;
       },
       style: (mvtFeature, tile) => {
+        const styleCacheKey = JSON.stringify(mvtFeature.properties);
```
keiya01, line 154: Properties in MVT feature is stringifyed json unless parsing properties, so I think it's almost no cost.

src/core/mantle/evaluator/simple/expression/expression.ts
```diff
@@ -24,11 +25,23 @@ export class Expression {
...
+    } else {
+      const originalExpression = expression;
+      expression = replaceDefines(expression, defines);
+      [expression, literalJP] = replaceVariables(
+        removeBackslashes(expression),
+        this._feature?.properties,
+        hasJSONPath,
+      );
+      REPLACED_VARIABLES_CACHE.set(originalExpression, [expression, literalJP]);
+    }
```
keiya01, line 44: I think we can cache replacedVariable without JSONPath. JSONPath is evaluated in replacedVariables, also it returns evaluated result, so we can not cache result value.

rot1024 (approved):

src/core/mantle/evaluator/simple/expression/expression.ts
```diff
@@ -24,11 +25,23 @@ export class Expression {
     this._expression = expression;
     this._feature = feature;
     let literalJP: JPLiteral[] = [];
-    expression = replaceDefines(expression, defines);
-    [expression, literalJP] = replaceVariables(
-      removeBackslashes(expression),
-      this._feature?.properties,
-    );
+
+    const hasJSONPath = /\$[a-zA-Z[.]/.test(expression);
```
pyshx, line 29:
```suggestion
    const hasJSONPath = /^\$([.](?!$)|\[[^\]]*])*$/.test(expression);
```

pyshx: The above regex will not be check statements like `$..book[?(@.price<10)]` because it contains the `..` operator and a filter expression `?(@.price<10)`

pyshx (approved):

src/core/mantle/evaluator/simple/expression/expression.ts
keiya01: I think we don't need to check entire expression because checking entire expression needs some extra cost like needing extra regexp steps.
So I think we can only check first two word.
JSON Path shold have ‘$‘ as first word. And second should be ‘.‘ or a-zA-Z.

pyshx: I see. You can merge this PR then.
Keep in mind that this simpler pattern may produce false positives, so you may need to further validate the matched strings to ensure that they are valid JSONPath expressions if that is a requirement for your use case.

keiya01: I rethought it based on your suggestion, I found a more efficient way. Could you check this commit?
[fix: use variable prefix](https://github.com/reearth/reearth-web/pull/574/commits/6872ec3ecb567f1a5a0cf9c3ff2a7798449b1297)

PR #590, opened 2023-03-30
fix: remove duplication of feature entity in reearth/core

pyshx:
# Overview
This PR removes duplication of entity of feature in cesium with the help a set.

### DIFF
--- a/src/core/engines/Cesium/Feature/index.tsx
+++ b/src/core/engines/Cesium/Feature/index.tsx
@@ -1,4 +1,4 @@
-import { DataType } from "@reearth/core/mantle";
+import { ComputedFeature, DataType } from "@reearth/core/mantle";
 import { getExtname } from "@reearth/util/path";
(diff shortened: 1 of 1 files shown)


src/core/engines/Cesium/Feature/index.tsx
```diff
@@ -84,6 +84,43 @@ export default function Feature({
     Array.isArray(displayType) &&
     displayType.every(k => components[k][1].noFeature && !components[k][1].noLayer);
 
+  const renderedComponents = new Set<string>(); // Initialize a set to store rendered components
+
+  const renderComponent = (k: keyof AppearanceTypes, f?: any) => {
```
rot1024, line 89: any -> ComputedFeature

pyshx: Yes, I'll keep that i mind🙏

PR #598, opened 2023-04-04
feat: add remembering last open workspace functionality

nina992:
# Overview
save last work space and show it when users open reearth next times


src/components/pages/Authentication/hooks.ts
```diff
@@ -14,23 +14,42 @@ export default () => {
...
   const passwordPolicy = window.REEARTH_CONFIG?.passwordPolicy;
   const t = useT();
 
   const { data, loading } = useGetTeamsQuery({ skip: !isAuthenticated });
-  const teamId = currentTeam?.id || data?.me?.myTeam.id;
+
+  if (isAuthenticated && !currentUserId) {
+    setCurrentUserId(data?.me?.id);
+  }
+  let teamId = currentTeam?.id || data?.me?.myTeam.id;
```
KaWaite, line 28: While you are making changes/updates, if there is a little bit of code you can clean up or fix up please do. We still have some "team" related code in reearth, but long ago we made "team" become "workspace". So this code here would be perfect to update to workspaceId

src/components/pages/Authentication/hooks.ts
KaWaite, line 1: @nina992 Can you take a little time to refactor this file? I feel like its getting messy, with too many variables/state that could be consolidated in to one (or anyways be simplified).
Maybe I am wrong, but please try to clean up the logic a bit

src/components/organisms/Settings/Workspace/hooks.ts
KaWaite, line 1: Same here as the other comment. Please try to clean up a bit

src/components/pages/Authentication/hooks.ts
```diff
@@ -5,32 +5,61 @@ import { useLocation, useNavigate } from "react-router-dom";
...
   const passwordPolicy = window.REEARTH_CONFIG?.passwordPolicy;
   const t = useT();
 
   const { data, loading } = useGetTeamsQuery({ skip: !isAuthenticated });
-  const teamId = currentTeam?.id || data?.me?.myTeam.id;
+
+  if (isAuthenticated && !currentUserId) {
+    setCurrentUserId(data?.me?.id);
+  }
+  const workspaceId = currentWorkspace?.id || data?.me?.myTeam.id;
```
KaWaite, line 27: I don't think we really need this since we can just use workspace.id.
But If you think this is necessary, please wrap it in a useMemo

src/components/pages/Authentication/hooks.ts
```diff
@@ -5,32 +5,61 @@ import { useLocation, useNavigate } from "react-router-dom";
...
   const [, setNotification] = useNotification();
   const passwordPolicy = window.REEARTH_CONFIG?.passwordPolicy;
   const t = useT();
 
   const { data, loading } = useGetTeamsQuery({ skip: !isAuthenticated });
-  const teamId = currentTeam?.id || data?.me?.myTeam.id;
+
+  if (isAuthenticated && !currentUserId) {
+    setCurrentUserId(data?.me?.id);
+  }
```
KaWaite, line 26: Can you be sure on first render that we will have values for `isAuthenticated` and `currentUserId`? If so, this is fine, but if not this should be in a useEffect so that WHEN we get those values we will check and setCurrentUserId if appropriate

src/components/pages/Authentication/hooks.ts
```diff
@@ -5,32 +5,61 @@ import { useLocation, useNavigate } from "react-router-dom";
...
+    [
+      currentUserId,
+      data?.me?.id,
+      data?.me?.teams,
+      data?.me?.myTeam,
+      setCurrentWorkspace,
+      navigate,
+      setCurrentUserId,
+    ],
+  );
```
KaWaite, line 58: Is there a difference you feel between the values such as data, currentUserId, etc and then workspaceId? Why i ask is because you are passing workspaceId as a prop, but the other values you use directly and add as deps.

src/components/pages/Authentication/hooks.ts
nina992: I think it is necessary especially first time for each user , Done for wrapping

src/components/pages/Authentication/hooks.ts
nina992: yes you are right , I handle it same as others

src/components/pages/Authentication/hooks.ts
nina992: I don't want this to be inside useeffect as it will be set for one time for each user on his browser cause the currentUser Id is also a atomWithLocalStorage and I don't want this to change once onther user use same browser

src/components/pages/Authentication/hooks.ts
nina992: 48b47ad76a7e35bbd1b45814258e6e94781b712f

src/components/pages/Authentication/hooks.ts
nina992: 48b47ad76a7e35bbd1b45814258e6e94781b712f

src/components/organisms/Authentication/RootPage/hooks.ts
```diff
@@ -1,24 +1,56 @@
...
+  if (isAuthenticated && !currentUserId) {
+    setCurrentUserId(data?.me?.id);
+  }
+
+  const workspaceId = useMemo(() => {
+    return currentWorkspace?.id || data?.me?.myTeam.id;
+  }, [currentWorkspace?.id, data?.me?.myTeam.id]);
+
+  const handleRedirect = useCallback(() => {
+    if (currentUserId == data?.me?.id) {
```
KaWaite, line 32: Please prefer `===` instead of `==` for the most part.

KaWaite (approved):

KaWaite (approved):

PR #599, opened 2023-04-05
feat: support styling color & show for 3dtiles model

airslice:
# Overview

3dtiles with no features are rendered as model, therefore the style of `coloe` `show` should also been included.

## What I've done

- Add `show` `color` to `MODEL_STYLE_PROPERTIES`
- Convert `show` `color` to match the type of style object for new Cesium3DTileStyle().

## How I tested

```js
reearth.layers.add({
  type: 'simple',
  data: {
    type: '3dtiles',
    url: "https://assets.cms.plateau.reearth.io/assets/0a/a2581a-e2cd-462a-a78b-a86841c477e9/uc22-024_sensynrobotics_kawasaki_build_2/tileset.json"
  },
  "3dtiles": {
    show: {
      "expression": {
        "conditions": [
          [
              "true",
              "true"
          ]
        ]
      }
    },
    color: {
      "expression": {
        "conditions": [
          [
              "true",
              "rgba(255,0,0,0.5)"
          ]
        ]
      }
    }
  }
})
```
## Screenshot

image

### DIFF
--- a/src/core/engines/Cesium/Feature/Tileset/hooks.ts
+++ b/src/core/engines/Cesium/Feature/Tileset/hooks.ts
@@ -95,13 +95,18 @@ type CachedFeature = {
 
 const MAX_NUMBER_OF_CONCURRENT_COMPUTING_FEATURES = 5000;
 
-type StyleProperty<N = string> = { name: N; convert?: "color" | "vec2" | "vec4" };
+type StyleProperty<N = string> = {
+  name: N;
+  convert?: "color" | "colorFunctionString" | "booleanString" | "vec2" | "vec4";
+};
 
 const COMMON_STYLE_PROPERTIES: StyleProperty<"color" | "show">[] = [
   { name: "color", convert: "color" },
   { name: "show" },
 ];
-const MODEL_STYLE_PROPERTIES: StyleProperty<"pointSize" | "meta">[] = [
+const MODEL_STYLE_PROPERTIES: StyleProperty<"color" | "show" | "pointSize" | "meta">[] = [
+  { name: "color", convert: "colorFunctionString" },
+  { name: "show", convert: "booleanString" },
   { name: "pointSize" },
   { name: "meta" },
 ];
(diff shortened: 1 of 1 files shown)


src/core/engines/Cesium/Feature/Tileset/hooks.ts
```diff
@@ -119,6 +124,10 @@ const TILESET_APPEARANCE_FIELDS: (keyof Cesium3DTilesAppearance)[] = [
 const convertStyle = (val: any, convert: StyleProperty["convert"]) => {
   if (convert === "color") {
     return toColor(val);
+  } else if (convert === "colorFunctionString") {
+    return `color("${val}")`;
+  } else if (convert === "booleanString") {
+    return typeof val === "boolean" ? (val ? "true" : "false") : val;
```
keiya01, line 130: Why does `show` need to be string?

src/core/engines/Cesium/Feature/Tileset/hooks.ts
```diff
@@ -95,13 +95,18 @@ type CachedFeature = {
...
+};
 
 const COMMON_STYLE_PROPERTIES: StyleProperty<"color" | "show">[] = [
   { name: "color", convert: "color" },
   { name: "show" },
 ];
-const MODEL_STYLE_PROPERTIES: StyleProperty<"pointSize" | "meta">[] = [
+const MODEL_STYLE_PROPERTIES: StyleProperty<"color" | "show" | "pointSize" | "meta">[] = [
+  { name: "color", convert: "colorFunctionString" },
+  { name: "show", convert: "booleanString" },
```
keiya01, line 109:
```suggestion
  { name: "color", convert: "color" },
  { name: "show" },
```

I think it should work, but is there problem?

src/core/engines/Cesium/Feature/Tileset/hooks.ts
airslice: I have updated this part removed the convert for boolean show.
I use this convert to avoid it been igored in `MODEL_STYLE_PROPERTIES.reduce`, there was `if(!val) return res;` and now i update it as `if (val === undefined) return res;` Is this okey? I'm not sure maybe i could miss something here.

src/core/engines/Cesium/Feature/Tileset/hooks.ts
airslice: Since this object is used as param of `new Cesium3DTileStyle` I think the `Color` type is not supported? If use Color cesium will crash with `TypeError: style.color.evaluateColor is not a function`

keiya01 (approved):

PR #608, opened 2023-04-12
perf: consider geojson as both delegate and active data type in reearth/core

pyshx:
# Overview
The PR adds support for large `geojson` type of data.
With the current approach of considering `geojson` only works to a limit of around 14k featureCollection in a layer, any data containing more features gives a WebGL error which basically means that `Cesium` cannot handle any more individual graphics component i.e `PolygonGraphics`, `PolylineGraphics` etc. 

# Description
As of now I've added a threshold limit of `6000` for featureCollection size, for geojson to be considered as an `active` data format in NLS and the size is large we consider it as a `delegate` data format.

Pros:

- We can now accommodate very large geojson with this approach now.
- We also save alot of overhead.

Cons:

- We lose the ability to custom styling on this data.
- With this approach, we can only handle featureCollection of size < 50k which is not bad but not that ideal.

## Side Note 
I've also utilised a LRU cache and `useMemo` to prevent re:rendering of component on camera movement, which has made things quite snappy for me, feel free to try that.

### DIFF
--- a/src/core/engines/Cesium/Feature/Resource/index.tsx
+++ b/src/core/engines/Cesium/Feature/Resource/index.tsx
@@ -2,7 +2,7 @@ import { Entity, type DataSource, Color } from "cesium";
 import { useCallback, useEffect, useMemo, useRef } from "react";
 import { KmlDataSource, CzmlDataSource, GeoJsonDataSource, useCesium } from "resium";
 
-import { ComputedFeature, DataType, evalFeature, Feature, guessType } from "@reearth/core/mantle";
+import { ComputedFeature, evalFeature, Feature, guessType } from "@reearth/core/mantle";
 import { requestIdleCallbackWithRequiredWork } from "@reearth/util/idle";
 
 import type { ResourceAppearance } from "../../..";
@@ -28,8 +28,6 @@ const comps = {
   czml: CzmlDataSource,
   geojson: GeoJsonDataSource,
 };
-
-const DataTypeListAllowsOnlyProperty: DataType[] = ["geojson"];
 
 type CachedFeature = {
   feature: Feature;
@@ -62,7 +60,7 @@ export default function Resource({
     const url = property?.url;
     return [
       type ?? (data?.type as ResourceAppearance["type"]),
-      url ?? (data && !DataTypeListAllowsOnlyProperty.includes(data.type) ? data?.url : undefined),
+      url ?? data?.url,
       !!data?.time?.updateClockOnLoad,
     ];
   }, [property, layer]);
@@ -132,6 +130,11 @@ export default function Resource({
     [updateClock, viewer?.clock],
   );
 
+  // convert hexCodeColorString to ColorValue?s
+  const strokeValue = stroke ? Color.fromCssColorString(stroke) : undefined;
+  const fillValue = fill ? Color.fromCssColorString(fill) : undefined;
+  const markerColorValue = markerColor ? Color.fromCssColorString(markerColor) : undefined;
+
   useEffect(() => {
     if (!viewer) return;
     cachedFeatures.current.forEach(f => {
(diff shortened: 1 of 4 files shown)


src/core/engines/Cesium/Feature/index.tsx
```diff
@@ -70,56 +73,85 @@ const pickProperty = (k: keyof AppearanceTypes, layer: ComputedLayer) => {
...
   layer,
   isHidden,
   ...props
 }: FeatureComponentProps): JSX.Element | null {
+  // Set the displayConfig based on the size of the feature collection array
+  if (layer.features?.length > FEATURE_DELEGATE_THRESHOLD) {
+    displayConfig.geojson = ["resource"];
+  } else {
+    displayConfig.geojson = "auto";
+  }
```
keiya01, line 89: I think it shouldn't override global variable, so we should define other local variable for this case.

src/core/engines/Cesium/Feature/index.tsx
```diff
@@ -70,56 +73,85 @@ const pickProperty = (k: keyof AppearanceTypes, layer: ComputedLayer) => {
...
+    const componentId = `${layer.id}_${f?.id ?? ""}_${k}`;
 
-  const renderedComponents = new Set<string>(); // Initialize a set to store rendered components
-
-  const renderComponent = (k: keyof AppearanceTypes, f?: ComputedFeature) => {
-    const [C, config] = components[k] ?? [];
-    if (!C || (f && !f[k]) || (config.noLayer && !f) || (config.noFeature && f)) {
-      return null;
+    // Check if the component output is already cached
+    const cachedComponent = CACHED_COMPONENTS.get(componentId);
```
keiya01, line 103: Question: If property is changed dynamically, is component re-rendered?

src/core/engines/Cesium/Feature/index.tsx
```diff
@@ -70,56 +73,85 @@ const pickProperty = (k: keyof AppearanceTypes, layer: ComputedLayer) => {
   return layer.layer[k];
 };
 
+const CACHED_COMPONENTS = new LRUCache<string, JSX.Element>({ max: 10000 });
```
keiya01, line 76: I think this LRUCache is instantiated by each layer. Doesn't this press memory?

src/core/engines/Cesium/Feature/Resource/index.tsx
```diff
@@ -132,6 +130,11 @@ export default function Resource({
     [updateClock, viewer?.clock],
   );
 
+  // convert hexCodeColorString to ColorValue?s
+  const strokeValue = stroke ? Color.fromCssColorString(stroke) : undefined;
+  const fillValue = fill ? Color.fromCssColorString(fill) : undefined;
+  const markerColorValue = markerColor ? Color.fromCssColorString(markerColor) : undefined;
```
keiya01, line 136: I think it's better to use `useMemo` because `Color.fromCssColorString` is a heavy task

src/core/engines/Cesium/Feature/index.tsx
pyshx: Good Catch. Fixed.

src/core/engines/Cesium/Feature/index.tsx
pyshx: It won't, I missed the property https://github.com/reearth/reearth-web/pull/608/commits/df1bdf392206201538d4d9cb4caaffde2361a790 should fix it.

src/core/engines/Cesium/Feature/index.tsx
pyshx: It does but the memory signature isn't that significant. If you want to we can reduce the maxSize.

src/core/engines/Cesium/Feature/Resource/index.tsx
pyshx: Done.

keiya01 (approved):
LGTM!

KaWaite (approved):

